PR #15, opened 2021-03-16
Exclude not to be tested lines from codecov

SimonBoothroyd:
## Description

This PR updates the codecov settings to not include lines which are not expected to be tested, including code within
a `__main__` statement, type checking imports, and `NotImplemented` exceptions.

This PR also adds a new pragma which can be used to explicitly state that blocks of code should be [excluded from the codecov counts](https://coverage.readthedocs.io/en/coverage-4.3.3/excluding.html#excluding-code-from-coverage-py). 

## Status
- [X] Ready to go

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -8,6 +8,13 @@ omit =
     # Omit generated versioneer
     openff/bespokefit/_version.py
 
+[coverage:report]
+exclude_lines =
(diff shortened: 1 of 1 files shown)


jthorton (approved):
Great idea!

PR #21, opened 2021-04-27
Migrate to new QCSubmit result collections

SimonBoothroyd:
## Description

This PR migrates this repo to use the new result collections introduced in https://github.com/openforcefield/openff-qcsubmit/pull/106. 

These objects greatly simplify the target factory code as i) result retrieval is now handled upstream and ii) they provide an almost identical API for consuming bespoke QC data or existing QC data.

The `ExistingQCData` object is removed as it is entirely replaced with the result collections.

The `test_optimizer_explicit` is also re-enabled here as the new result collections allowed the target creation code to consume the bespoke QC data almost with no code changes.

## Status
- [X] Ready to go


openff/bespokefit/schema/data.py
```diff
@@ -178,42 +182,21 @@ def update_with_results(
         if len(self.tasks) == 0:
             return
 
-        expected_task_type = self.tasks[0].task_type
```
SimonBoothroyd, line 181: The type checks are now handled by the task `update_with_results` functions

openff/bespokefit/tests/schema/bespoke/test_tasks.py
```diff
@@ -177,146 +187,37 @@ def test_fitting_entry_not_equal(ethane, ethane_opt_task):
...
-    """
-
-    schema = get_fitting_schema(occo)
-
-    results = TorsionDriveCollectionResult.parse_file(
-        get_data_file_path(os.path.join("test", "qc-datasets", "occo", "occo.json"))
-    )
-
-    with pytest.raises(Exception):
-        schema.update_with_results(results=results)
```
SimonBoothroyd, line 322: Moved to test_fitting.py

PR #24, opened 2021-04-27
Fix ordering of grid ids in torsion drive (FB) metadata

SimonBoothroyd:
## Description

This PR fixes a potential bug whereby the grid ids included in the `metadata.json` prodcued by the force balance `TorsionProfileTargetFactory` were ordered differently then in `qdata.txt` and `scan.xyz`.

This PR also adds a missed `NotImplementedError` when trying to fit forces using the force balance ab initio target.

## Status
- [X] Ready to go


SimonBoothroyd:
The test failures are unrelated and should be fixed by #24 so I'll go ahead and merge this.

PR #36, opened 2021-05-12
Add initial `openff-bespoke prepare` CLI

SimonBoothroyd:
## Description

This PR exposes a new `openff-bespoke prepare` CLI:

```
Usage: openff-bespoke prepare [OPTIONS]

  Prepare an optimizations' input files based on its schema.

Options:
  -i, --input FILE  The file path to a JSON serialized optimization schema.
                    This can be either a general or a bespoke schema.

  --help            Show this message and exit.
```

which is built around the `click` package. Currently if a `bespoke` schema is used as input no QC is spun up and generated when missing, although in future we may consider adding this.

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/cli/prepare.py
+++ b/openff/bespokefit/cli/prepare.py
@@ -1,0 +1,62 @@
+import json
+
+import click
+from click import ClickException
+from pydantic import ValidationError
+
+from openff.bespokefit.optimizers import get_optimizer
+from openff.bespokefit.schema.fitting import (
+    BespokeOptimizationSchema,
+    OptimizationSchema,
+)
+
+_SCHEMA_TYPES = {"general": OptimizationSchema, "bespoke": BespokeOptimizationSchema}
+
+
+@click.command("prepare")
+@click.option(
+    "-i",
+    "--input",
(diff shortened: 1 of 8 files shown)


openff/bespokefit/cli/prepare.py
```diff
@@ -0,0 +1,62 @@
...
+from openff.bespokefit.schema.fitting import (
+    BespokeOptimizationSchema,
+    OptimizationSchema,
+)
+
+_SCHEMA_TYPES = {"general": OptimizationSchema, "bespoke": BespokeOptimizationSchema}
+
+
+@click.command("prepare")
+@click.option(
```
jthorton, line 17: Should this be an argument as it is required to run? This would also make the command slightly simpler as we would not need the `-i` flag?

jthorton (approved):
This is a great start to the CLI and LGTM I have one minor point but it's not blocking.

SimonBoothroyd:
Thanks! For now i've left it as is in case we want to allow users to specify another type of input path (e.g. a url, a RESTFul API, etc), however happy to revisit it later!

PR #52, opened 2021-09-17
Add support for supplying local QC data

SimonBoothroyd:
## Description

This PR adds a new `LocalQCData` type that allows users to pass QC data generated using QCEngine directly to an input schema.

## Status
- [X] Ready to go


jthorton (approved):
LGTM! One step closer to local runs!

PR #76, opened 2021-10-08
Add initial support for multi-stage optimizations

SimonBoothroyd:
## Description

This PR handles the initial model reshaping that will allow this framework to perform multi-stage optimisations, for example `fit charges -> find bond + angles -> fit torsions`, seamlessly.

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/qcgenerator/worker.py
+++ b/openff/bespokefit/executor/services/qcgenerator/worker.py
@@ -102,7 +102,6 @@ def compute_torsion_drive(task_json: str) -> TorsionDriveResult:
         ),
     )
 
-
     return_value = qcengine.compute_procedure(
         input_schema, "torsiondrive", raise_error=True
     )


openff/bespokefit/executor/services/optimizer/worker.py
```diff
@@ -20,12 +21,34 @@
...
     )
     input_schema.id = self.request.id
 
-    optimizer = get_optimizer(input_schema.optimizer.type)
-    result = optimizer.optimize(input_schema, keep_files=False)
+    input_force_field = ForceField(input_schema.initial_force_field)
+
+    stage_results = []
+
+    for stage in input_schema.stages:
```
jthorton, line 35: I might have missed if this is handled in the coordinator somewhere but I think we should just be expecting a `OptimizationStageSchema` and perform one optimisation? or is the idea to do all tasks for that stage at once? So when the schema gets here all stages have already had their reference data generated? If so I think the qcgenerator /fragmentation stages will need to be updated to do all tasks for every optimisation stage as well.

PR #77, opened 2021-10-13
Bespoke clean up

jthorton:
## Description
This PR fixes some small issues in the bespokefit pipeline and simplifies some of the models.

## Todos
  - [x] Remove the bespoke workflow factory smirks argument as this is included in the parameter hyper parameters field
  - [x] add the option to keep fitting files
  - [x] Remove the optimisation schema and re-use the qcsubmit object
  - [x] Fix default geometric settings in torsion drives
  - [x] Make sure we try and select heavy atoms in torsiondrives where possible
  - [x] Fix the initial values for bespoke torsion parameters
  - [x] Fix `initial_parameter_values` to return all of the fitting attributes per parameter

## Questions
- [ ] How should we expose the keep files option? as a workflow factory setting?

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/optimizer/worker.py
+++ b/openff/bespokefit/executor/services/optimizer/worker.py
@@ -26,6 +26,6 @@ def optimize(self, optimization_input_json: str) -> str:
     input_schema.id = self.request.id
 
     optimizer = get_optimizer(input_schema.optimizer.type)
-    result = optimizer.optimize(input_schema, keep_files=input_schema.keep_files)
+    result = optimizer.optimize(input_schema, keep_files=settings.BEFLOW_KEEP_FILES)
 
     return serialize(result, encoding="json")
--- a/openff/bespokefit/executor/services/settings.py
+++ b/openff/bespokefit/executor/services/settings.py
@@ -35,3 +35,4 @@ class Settings(BaseSettings):
     BEFLOW_OPTIMIZER_PREFIX = "optimizations"
     BEFLOW_OPTIMIZER_ROUTER = "openff.bespokefit.executor.services.optimizer.app:router"
     BEFLOW_OPTIMIZER_WORKER = "openff.bespokefit.executor.services.optimizer.worker"
+    BEFLOW_KEEP_FILES = False
--- a/openff/bespokefit/schema/fitting.py
+++ b/openff/bespokefit/schema/fitting.py
@@ -60,11 +60,6 @@ class BaseOptimizationSchema(SchemaBase, abc.ABC):
         description="The fittings targets to simultaneously optimize against.",
     )
 
-    keep_files: bool = Field(
-        True,
-        description="If the files generated during a parameter optimization should be kept or not.",
-    )
-
     @property
     def n_targets(self) -> int:
         """Returns the number of targets to be fit."""
--- a/openff/bespokefit/tests/executor/services/qcgenerator/test_worker.py
+++ b/openff/bespokefit/tests/executor/services/qcgenerator/test_worker.py
@@ -1,6 +1,5 @@
 import json
 
-import pytest
 from openff.toolkit.topology import Molecule
 from qcelemental.models.common_models import Model
 from qcelemental.models.procedures import OptimizationResult, TorsionDriveResult


openff/bespokefit/workflows/bespoke.py
```diff
@@ -209,14 +203,16 @@ def _pre_run_check(self) -> None:
...
-                "selected to be optimized."
-            )
         else:
             return
 
+    @property
+    def target_smirks(self) -> List[SMIRKSType]:
+        """Returns a list of the target smirks types based on the selected hyper parameters."""
+        return list(
+            {SMIRKSType(parameter.type) for parameter in self.parameter_hyperparameters}
```
SimonBoothroyd, line 213: Great suggestion!

openff/bespokefit/utilities/smirks.py
```diff
@@ -256,15 +256,29 @@ def generate_smirks_from_fragment(
...
+                    default_k[p - 1] = parameter.k[i]
+
+                # update with new parameters
+                parameter.k = default_k
+                parameter.phase = default_phase
+                parameter.periodicity = default_p
+
+                # TODO make sure we do not fit interpolated parameters
+                # if parameter.k_bondorder is not None:
+                #     parameter.k_bondorder.append(parameter.k_bondorder[i - 1])
```
SimonBoothroyd, line 281: Should we maybe raise `NotImplementedError` here just in case someone tries to slip it through?

openff/bespokefit/schema/tasks.py
```diff
@@ -84,15 +60,12 @@ class Torsion1DTaskSpec(QCGenerationTask):
...
+    scan_range: Optional[Tuple[int, int]] = Field(
+        None, description="The range of grid angles to scan."
     )
-
-    program: str = Field(..., description="The program to use to evaluate the model.")
-    model: Model = Field(..., description=str(Model.__doc__))
-
-    optimization_spec: OptimizationSpec = Field(
-        OptimizationSpec(),
+    # update to use the default torsiondrive settings
```
SimonBoothroyd, line 66: Is this a TODO?

openff/bespokefit/schema/tasks.py
```diff
@@ -54,22 +43,9 @@ class HessianTask(HessianTaskSpec):
...
 
 
-class OptimizationTaskSpec(QCGenerationTask):
+class OptimizationTaskSpec(HessianTaskSpec):
     type: Literal["optimization"] = "optimization"
 
-    n_conformers: conint(gt=0) = Field(
-        ...,
-        description="The maximum number of conformers to begin the optimization from.",
-    )
```
SimonBoothroyd, line 63: Do we use `n_conformers` in the QC generation stages?

openff/bespokefit/schema/tasks.py
```diff
@@ -1,28 +1,21 @@
...
-
-
 class QCGenerationTask(BaseModel, abc.ABC):
 
     type: Literal["base-task"]
 
+    program: str = Field(..., description="The program to use to evaluate the model.")
+    model: Model = Field(..., description=str(Model.__doc__))
+
```
SimonBoothroyd, line 18: Did you want to replace these with the QCSpec model at some point?

openff/bespokefit/executor/services/qcgenerator/worker.py
```diff
@@ -75,10 +93,9 @@ def compute_torsion_drive(task_json: str) -> TorsionDriveResult:
             model=task.model, driver=DriverEnum.gradient
         ),
         optimization_spec=OptimizationSpecification(
-            procedure=task.optimization_spec.procedure,
+            procedure="geometric",
```
SimonBoothroyd, line 96:
```suggestion
            procedure=task.optimization_spec.program,
```
I think this is what is used further down?

jthorton: Yes that is better this should allow use to easily swap out the optimiser in future as well!

openff/bespokefit/schema/tasks.py
jthorton: Yes I think the validation in the QCSpec model should help avoid a lot of errors Ill swap these fields out for a `qc_spec` field.

jthorton: This is done in [df5a6bf](https://github.com/openforcefield/bespoke-fit/pull/77/commits/df5a6bfdab02fc000479ab3c405e944b1e16744a) but this means that the qc cache now depends on arbitrary fields like the spec name and spec description, maybe the QCSpec object should be split in qcsubmit to offer a version without these fields but which still offers the validation?

openff/bespokefit/schema/tasks.py
SimonBoothroyd: Oh ha I totally missed that - that's great then!

PR #81, opened 2021-10-29
Fragmentation cache

jthorton:
## Description
This PR adds a simple fragmentation cache based on the fixed hydrogen inchi key of the molecule meaning it should be independent of the ordering of the molecule. 

## Status
- [x] Ready to go


SimonBoothroyd (approved):
LGTM!

PR #87, opened 2021-11-09
Move parameter generation to the optimisation stage

jthorton:
## Description
This PR reworks the parameter generation step into the optimisation stage. The bespoke parameter generation settings are now passed to the `BespokeOptimizationSchema` and are used to generate the new smirks patterns. The rest of the workflow is the same. 

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -165,7 +165,9 @@ class QCGenerationStage(_Stage):
             ),
             None,
         )
-        fragments = [] if fragment_stage is None else fragment_stage.result.fragments
+        fragments = (
+            [] if fragment_stage.result is None else fragment_stage.result.fragments
+        )
 
         target_qc_tasks = defaultdict(list)
 
@@ -313,8 +315,8 @@ class OptimizationStage(_Stage):
 
     @staticmethod
     async def _generate_parameters(
-        fragmentation_result: FragmentationResult,
         input_schema: BespokeOptimizationSchema,
+        fragmentation_result: Optional[FragmentationResult],
     ):
         """
         Generate a list of parameters which are to be optimised, these are added to the input force field.
@@ -402,7 +404,10 @@ class OptimizationStage(_Stage):
 
         # TODO: Move these methods onto the celery worker.
         try:
-            await self._generate_parameters(fragmentation_stage.result, input_schema)
+            await self._generate_parameters(
+                fragmentation_result=fragmentation_stage.result,
+                input_schema=input_schema,
+            )
         except BaseException as e:  # lgtm [py/catch-base-exception]
 
             self.status = "errored"
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -5,7 +5,14 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Union
 
 import httpx
 from openff.fragmenter.fragment import FragmentationResult
-from openff.toolkit.typing.engines.smirnoff import ParameterType
+from openff.toolkit.typing.engines.smirnoff import (
+    AngleHandler,
+    BondHandler,
+    ImproperTorsionHandler,
+    ParameterType,
+    ProperTorsionHandler,
+    vdWHandler,
+)
 from pydantic import Field
 from qcelemental.models import AtomicResult, OptimizationResult
 from qcelemental.util import serialize
@@ -32,10 +39,20 @@ from openff.bespokefit.executor.utilities.typing import Status
 from openff.bespokefit.schema.data import BespokeQCData, LocalQCData
 from openff.bespokefit.schema.fitting import BespokeOptimizationSchema
 from openff.bespokefit.schema.results import BespokeOptimizationResults
-from openff.bespokefit.schema.smirnoff import ProperTorsionSMIRKS
+from openff.bespokefit.schema.smirnoff import (
+    AngleSMIRKS,
+    BondSMIRKS,
+    ImproperTorsionSMIRKS,
+    ProperTorsionSMIRKS,
+    VdWSMIRKS,
+)
 from openff.bespokefit.schema.tasks import Torsion1DTask
 from openff.bespokefit.utilities.pydantic import BaseModel
-from openff.bespokefit.utilities.smirks import ForceFieldEditor, SMIRKSGenerator
+from openff.bespokefit.utilities.smirks import (
+    ForceFieldEditor,
+    SMIRKSGenerator,
+    SMIRKSType,
+)
 
 if TYPE_CHECKING:
     from openff.bespokefit.executor.services.coordinator.models import CoordinatorTask
@@ -268,13 +285,18 @@ class OptimizationStage(_Stage):
     )
 
     @staticmethod
-    def _regenerate_torsion_parameters(
+    def _generate_torsion_parameters(
         fragmentation_result: FragmentationResult,
-        initial_force_field: str,
+        input_schema: BespokeOptimizationSchema,
     ) -> List[ParameterType]:
 
         parent = fragmentation_result.parent_molecule
-        smirks_gen = SMIRKSGenerator(initial_force_field=initial_force_field)
+        smirks_gen = SMIRKSGenerator(
+            initial_force_field=input_schema.initial_force_field,
+            generate_bespoke_terms=input_schema.generate_bespoke_terms,
+            expand_torsion_terms=input_schema.expand_torsion_terms,
+            target_smirks=[SMIRKSType.ProperTorsions],
+        )
 
         new_smirks = []
         for fragment_data in fragmentation_result.fragments:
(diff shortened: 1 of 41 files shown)


openff/bespokefit/schema/smirnoff.py
```diff
@@ -95,8 +95,8 @@ class VdWHyperparameters(BaseSMIRKSHyperparameters):
 
     type: Literal["vdW"] = "vdW"
 
-    priors: Dict[Literal["epsilon", "rmin_half"], PositiveFloat] = Field(
-        {"epsilon": 0.1, "rmin_half": 0.1}, description=""
```
SimonBoothroyd, line 99: Is there a reason to use `sigma` over `rmin_half` here?

openff/bespokefit/utilities/smirks.py
```diff
@@ -335,12 +335,17 @@ def _get_bespoke_parameters(
...
+                self._get_bespoke_smirks(
+                    parent=parent,
+                    fragment=fragment,
+                    fragment_map_indices=fragment_map_indices,
+                    fragment_is_parent=fragment_is_parent,
+                    smirks_type=smirk_type,
+                )
             )
-            for smirk_type in self.target_smirks
-        }
```
SimonBoothroyd, line 343: Hmm I wonder if this will cause us issues when we come to fit say BCCs or V-Sites that have two parent atoms as these will on the surface look like bond SMIRKS?

openff/bespokefit/workflows/bespoke.py
```diff
@@ -415,88 +398,88 @@ def _optimization_schema_from_records(
...
+        #
+        #     dihedrals = record.keywords.dihedrals
+        #     assert len(dihedrals) == 1, "only 1D torsion drives are supported"
+        #
+        #     bespoke_parameters.extend(
+        #         smirks_gen.generate_smirks_from_molecule(
+        #             molecule=record_molecule,
+        #             central_bond=(dihedrals[0][1], dihedrals[0][2]),
+        #         )
+        #     )
```
SimonBoothroyd, line 417: Hmm I think in the world where we don't provide explicit parameters we'd need the fragmentation step to consider what local QC data has been provided and don't try and generate new fragments for the existing torsion drives?

openff/bespokefit/schema/fitting.py
```diff
@@ -81,6 +83,19 @@ class BaseOptimizationSchema(SchemaBase, abc.ABC):
...
+            {
+                SMIRKSType(parameter.type)
+                for stage in self.stages
+                for parameter in stage.parameter_hyperparameters
+            }
+        )
+
     @property
     def initial_parameter_values(
         self,
```
SimonBoothroyd, line 101: This maybe needs moving to the general schema as it will always be empty for the bespoke one?

openff/bespokefit/schema/fitting.py
```diff
@@ -81,6 +83,19 @@ class BaseOptimizationSchema(SchemaBase, abc.ABC):
...
+        Used to determine which parameters should be fit.
+        """
+        return list(
+            {
+                SMIRKSType(parameter.type)
+                for stage in self.stages
+                for parameter in stage.parameter_hyperparameters
+            }
+        )
+
```
SimonBoothroyd, line 98: It may be better to have this live on the bespoke schema as I don't think we use this in the general fits? We can always easily move it back here to the parent class later if we need it without breaking the API.

openff/bespokefit/schema/fitting.py
```diff
@@ -138,3 +153,18 @@ class BespokeOptimizationSchema(BaseOptimizationSchema):
...
+    expand_torsion_terms: bool = Field(
+        True,
+        description="If the optimization should first expand the number of k values "
+        "that should be fit for each torsion beyond what is in the initial force field.",
+    )
+    generate_bespoke_terms: bool = Field(
+        True,
+        description="If the optimized smirks should be bespoke to the target molecules.",
+    )
```
SimonBoothroyd, line 165: I wonder if we should maybe wrap these in a new `BespokeSMIRKS` class (or something similar, I can't think of a good name right now!), and then move them directly into the optimization stage model so `parameters` becomes a union of `List[SMIRNOFFParameter]` and `BespokeSMIRKS`? That way we don't have an empty `parameters` field in addition to this other fields, and it also means each stage has more control over how the parameters are generated. E.g. you might want bespoke torsions but maybe the exact AM1BCCs.

openff/bespokefit/executor/services/coordinator/stages.py
```diff
@@ -290,44 +312,64 @@ def _regenerate_torsion_parameters(
...
-                new_parameters.append(old_parameter)
-
-        # now add our new parameters to the force field and the schema list
-        for torsion_parameter in torsion_parameters:
-            new_parameters.append(
-                ProperTorsionSMIRKS(
-                    smirks=torsion_parameter.smirks, attributes={"k1", "k2", "k3", "k4"}
-                )
+            parameters = smirks_gen.generate_smirks_from_molecule(
+                molecule=input_schema.molecule
```
SimonBoothroyd, line 346: Should this logic be based on the QC data that's available rather than fragments? This way it should capture parameters for the new fragments as well as for any user provided QC data

SimonBoothroyd (commented):
Looking great so far! I had a couple of questions that would be great to get your thoughts on!

openff/bespokefit/schema/smirnoff.py
jthorton: This was just to align the `VdWSMIRKS` and `VdWHyperparameters` objects to use the same naming happy to go with either though if you would prefer `rmin_half`?

openff/bespokefit/utilities/smirks.py
jthorton: Good point and this will currently cause an issue with improper vs proper torsions, my other idea was to return the new SMIRKS as [SMIRNOFFParameter](https://github.com/openforcefield/bespoke-fit/blob/cd9749318f73b57494b65a2dd2af9035ccec4e59/openff/bespokefit/schema/smirnoff.py#L257) where the types can be used to avoid this problem.

openff/bespokefit/schema/fitting.py
jthorton: Yeah it's empty until we are inside the workflow. It could be useful if users want to compare the initial and final parameters in the fitting result?

openff/bespokefit/schema/fitting.py
jthorton: Great idea, I did also toy around with putting these settings in the hyperparameter objects but maybe that is too much control and we can just start with the optimization stage model. Maybe something like `SMIRKSettings` and I think I would prefer the settings to have their own field to keep things organised, validation would also be easier if the field is required rather than looping over the list to make sure it has been supplied.

jthorton: > That way we don't have an empty `parameters` field in


In most cases, the `parameters` field would be empty as we plan on generating these later in the workflow, but we could also allow users to pass their own list of parameters they wish to be optimised for example when they want to fit a subset of bonds and angles and not all of them which is the default behaviour. In this case, if the user passes the `SMIRKSettings.generate_bespoke_terms=False` and a list of parameters we would respect them and only fit those?

jthorton:
## Update
I have added the ability for the fragmentation engine to be `None` which allows users to avoid this stage and fragments are mocked based on the target torsion smirks and if we are fitting torsions or not. This now means that we have a consistent logic flow based on the fragmentation stage which determines the required smirks, torsiondrives and optimisations. 

To avoid the issues with supplying QC data or generating tasks from this data, raised by @SimonBoothroyd above we aim to remove this in the next PR which will allow users to supply QCArchive data as a local cache supplement only. This will allow users to create optimisations from results records by setting fragmentation to `None` and providing torsion target smirks which only hit known driven bonds.

openff/bespokefit/executor/services/fragmenter/cache.py
```diff
@@ -27,7 +31,7 @@ def cached_fragmentation_task(
 
     task_id = worker.fragment.delay(
         cmiles=task.cmiles,
-        fragmenter_json=task.fragmenter.json(),
+        fragmenter_json=fragment_string,
         target_bond_smarts=task.target_bond_smarts,
```
SimonBoothroyd, line 35:
```suggestion
        target_bond_smarts=target_bonds_string,
```

openff/bespokefit/executor/services/optimizer/worker.py
```diff
@@ -26,29 +29,35 @@ def optimize(self, optimization_input_json: str) -> str:
...
-                result.refit_force_field, allow_cosmetic_attributes=True
-            ).to_string(discard_cosmetic_attributes=True)
-        )
+            input_force_field = ForceField(
+                ForceField(
+                    result.refit_force_field, allow_cosmetic_attributes=True
+                ).to_string(discard_cosmetic_attributes=True)
+            )
+    if not settings.BEFLOW_KEEP_FILES:
+        shutil.rmtree(input_schema.id, ignore_errors=True)
```
SimonBoothroyd, line 60: Hmm we may need to be a bit careful here - tasks should be idempotent, which basically means you should be able to run the same task twice and not have it interfere with itself.

To ensure this is the case, we should maybe run the optimization in a temp directory, and then copy it back to a non-temp directory when it's finished if keep files is true?

openff/bespokefit/schema/fitting.py
```diff
@@ -138,3 +140,26 @@ class BespokeOptimizationSchema(BaseOptimizationSchema):
         "By default bespoke torsion parameters (if requested) will be constructed for "
         "all non-terminal 'rotatable bonds'",
     )
+
+    smirk_settings: SMIRKSettings = Field(
+        SMIRKSettings(),
+        description="The settings that should be used when generating SMIRKS patterns for this optimization stage.",
+    )
```
SimonBoothroyd, line 147: Do we maybe need to split this class into a bespoke and general version like we do with the fitting schemas? This seems like a field that should exist on a bespoke schema but probably not a general one!

openff/bespokefit/schema/smirnoff.py
SimonBoothroyd: Ah that makes sense! Probably best to keep the two aligned!

openff/bespokefit/utilities/smirks.py
```diff
@@ -132,43 +133,52 @@ def make_smirks_attribute_graph(chem_env: ChemicalEnvironment) -> nx.Graph:
...
+#     tagged_atoms = len(smirk.get_indexed_atoms())
+#
+#     assert tagged_atoms == expected_tags, (
+#         f"The smirks pattern ({smirks}) has {tagged_atoms} tagged atoms, but should "
+#         f"have {expected_tags}."
+#     )
+#
+#     return smirks
+
+
```
SimonBoothroyd, line 151:
```suggestion
```

openff/bespokefit/workflows/bespoke.py
```diff
@@ -415,88 +405,88 @@ def _optimization_schema_from_records(
...
+    #     """
+    #     smirks_gen = SMIRKSGenerator(
+    #         initial_force_field=self.initial_force_field,
+    #         generate_bespoke_terms=self.generate_bespoke_terms,
+    #         expand_torsion_terms=self.expand_torsion_terms,
+    #         target_smirks=self.target_smirks
+    #         if target_smirks is None
+    #         else target_smirks,
+    #     )
+    #     return smirks_gen
```
SimonBoothroyd, line 489: Is this safe to just delete?

openff/bespokefit/workflows/bespoke.py
```diff
@@ -513,30 +503,30 @@ def _build_optimization_schema(
         self,
         molecule: Molecule,
         index: int,
-        bespoke_parameters: Optional[List[ParameterType]] = None,
+        # bespoke_parameters: Optional[List[ParameterType]] = None,
```
SimonBoothroyd, line 506:
```suggestion
```

SimonBoothroyd (approved):
A few minor comments but otherwise I think this LGTM! Thanks for doing this overhaul!

openff/bespokefit/executor/services/fragmenter/cache.py
jthorton: I am not sure this works if `target_bonds_string` is `null` or should I be using json loads and dumps here?

openff/bespokefit/schema/fitting.py
jthorton: Currently, this is only on the `BespokeOptimizationSchema` and I don't think we have any options we want to expose like this to the general schema yet? I think for now it might be best to only keep this on the bespoke model and we can update the general one later if we need to.

PR #91, opened 2021-11-10
Make the CLI force field argument optional

SimonBoothroyd:
## Description

This PR makes the force field argument passed to the executor CLI, such that it will only overwrite the schema value if provided.

## Status
- [X] Ready to go


jthorton (approved):
LGTM!

PR #92, opened 2021-11-10
Add clearer error messages to executor stages

SimonBoothroyd:
## Description

This PR adds clearer error messages for when things go wrong during the executor stages including:

* no fragments were generated for a given molecule
* for some reason no QC data was generated that matches a given target
* an uncaught error is raised

## Status
- [X] Ready to go


jthorton (commented):
LGTM!

PR #106, opened 2021-12-02
cache update via cli

jthorton:
## Description
This PR adds the ability to update the QC calculation cache using results computed in qcarchive, these are converted to local results and mock celery results are added to a local redis instance.

## Todos
  - [ ] support qcsubmit results files
  - [ ] support qcarchive directly 
  - [ ] add tests

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/cli/cache.py
+++ b/openff/bespokefit/cli/cache.py
@@ -1,47 +1,195 @@
+import hashlib
+import json
+import uuid
+from typing import Optional, Union
+
 import click
 import redis
 import rich
 from click_option_group import optgroup
+from openff.qcsubmit.results import (
+    OptimizationResultCollection,
+    TorsionDriveResultCollection,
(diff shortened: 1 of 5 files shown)


openff/bespokefit/cli/cache.py
```diff
@@ -0,0 +1,195 @@
...
+
+    pretty.install()
+    console = rich.get_console()
+    print_header(console)
+
+    console.print(Padding("1. gathering QCSubmit results", (0, 0, 1, 0)))
+
+    with console.status("loading results file"):
+        try:
+            qcsubmit_result = TorsionDriveResultCollection.parse_file(input_file_path)
```
jthorton, line 70: @SimonBoothroyd I think this could be simplified if we a type field to the collection objects in qcsubmit.

PR #109, opened 2021-12-16
Stop celery result deletion

jthorton:
## Description
This PR should stop saved celery results from being deleted after 1 day they should now be saved indefinitely.  Previously this would cause any cached stages to hang when trying to find the result in redis after it had been removed. 

## Status
- [x] Ready to go


SimonBoothroyd:
Ooof great catch!

PR #115, opened 2022-02-10
Extend the getting started docs and slightly re-organise

SimonBoothroyd:
## Description

This PR aims to extend the getting started docs to emphasise that for most users the CLI is the way to interface with the framework given this is mostly industry targeted. It also re-organises the layout slightly to try and make the user docs a bit less developer focused.

I didn't have time to go over the executor and workflow docs, but from a quick scan they should be good for a first pass.

## Status
- [X] Ready to go


docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,286 @@
...
+        results = wait_until_complete(task.id).results
+    
+    # Print out the resulting force field in OFFXML format
+    print(results.stages[-1].refit_force_field)
+```
+
+The `BespokeExecutor` not only takes care of calling out to any external programs in your workflow such as when 
+generating reference QC data, it also manages spreading a queue of tasks over a pool of worker threads so that fitting 
+can be executed efficiently in parallel. The `BespokeExecutor` is described in more detail in 
+[it's own chapter](executor_chapter).
```
Yoshanuikabundi, line 208:
```suggestion
[its own chapter](executor_chapter).
```

docs/users/bespoke-workflows.md
```diff
@@ -1,18 +1,22 @@
...
+
 ## The default workflow
 
-The default workflow factory is suitable for adding bespoke torsion terms to a
-molecule for the [Parsley Open Force Field]. In this section, we'll build this
-workflow factory up from nothing, but this is just for a demonstration; you can
-always build this factory by simply instantiating the `BespokeWorkflowFactory`
-class without arguments.
+The default workflow is suitable for augmenting a general [SMIRNOFF] force field (currently the 
+[Parsley Open Force Field]) with a set of bespoke torsion terms trained to reproduce a QC torsion scan.
```
Yoshanuikabundi, line 16:
```suggestion
The default workflow is suitable for augmenting a general [SMIRNOFF] force field (currently the 
[Parsley Open Force Field]) with a set of bespoke torsion terms trained to reproduce a QC torsion scan. 

[SMIRNOFF]: https://openforcefield.github.io/standards/standards/smirnoff/
```

The SMIRNOFF link needs to be defined somewhere.

docs/users/bespoke-executor.md
```diff
@@ -153,10 +153,10 @@ explicitly or implicitly.
 Molecules may also be submitted from a chemical structure file:
 
 ```sh
-openff-bespoke executor submit --input molecule.sdf
+openff-bespoke executor submit --file molecule.sdf
```
Yoshanuikabundi, line 156: I'll see what I can do about adding tests for shell commands to the testing framework

docs/index.md
```diff
@@ -47,13 +76,24 @@ so we must use eval-rst here
...
+caption: "Developer Guide"
+glob: True
+---
 
-* [](genindex)
-* [](modindex)
-* [](search)
+developers/schemas
+
+:::
```
Yoshanuikabundi, line 99: I resisted writing up Pydantic schemas for a while until I decided that a basic understanding is pretty essential to understanding the Python API - jargon like fields and models and schemas are all over the API reference, and thinking of some python types as just data (as opposed to stateful behaviour) isn't necessarily totally obvious to Pythonistas. So I'd suggest either moving the schemas chapter back into the user's guide. I'm not sure the schemas chapter is really enough details for developers of bespoke-fit, who would be better served by the Pydantic docs.

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,286 @@
...
+//          basis=None,
+//          program="xtb",
+//          spec_name="xtb",
+//          spec_description="gfn2xtb",
+//      )
+//  ]
+//  input_molecule = Molecule.from_smiles("C(C(=O)O)N") # Glycine  
+//  workflow_schema = factory.optimization_schema_from_molecule(
+//      molecule=input_molecule
+//  )
```
Yoshanuikabundi, line 188: What did you think of this style of writing doctests? Do you think we'd be better off just using notebooks, at least for these examples where we're essentially breaking one python session into multiple code examples?

Yoshanuikabundi (approved):
These changes generally seem like an improvement - sorry I didn't realise the focus on the CLI. I found one or two typos and things I'd change.

docs/getting-started/quick-start.md
SimonBoothroyd: Hmm I'd never used this syntax before so it was a little bit foreign to me. My main complaint is that PyCharm does not understand the `//`, and so things like code-completion don't work as well in cells that contain these tests and in general the IDE gets lit up with quite a few errors / warnings that make real errors / warnings harder to spot!

Maybe a notebook would be a better way to go. We can maybe make the switch in a future PR?

docs/index.md
SimonBoothroyd: Hmm that's a good point. We may want to go over the API docs themselves to try and make them more clear. Terms like *field* (sometimes called *attributes* or *properties*), *model*, and *schema* are I think quite common terms in programming, and are agnostic of `pydantic`, but we may want to try and make their usage a little more consistent.

Because this package is predominately industry facing, I think in general most users will be more inclined to use the CLI as that is how they typically seem to interact with their tools. For those who do write python, I imagine they will mostly be copying the code examples and maybe making slight edits based on the text in the main user docs rather than reading the API docs, so I think it's fine for the API docs to be a little more technical / jargony 🙂

docs/getting-started/quick-start.md
SimonBoothroyd: Maybe we could just use [`testsetup`](https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html#directive-testsetup) blocks instead?

PR #116, opened 2022-02-11
Add dedicated executor output model

SimonBoothroyd:
## Description

This PR adds a new `BespokeExecutorOutput` model that will store any outputs from the bespoke executor. This is used in place of the more abstract `CoordinatorGETResponse` which is more of an internal model, and exposes a number of QoL features such as getting the current status, error message, or final force field.

Further, this PR changes the way that errors are handled by the executor such that the executor raises exceptions directly when running via the python API, but that are captured and turned into more understandable versions when running through the CLI. This is made possible by a new context manager `handle_common_errors`, e.g.:

```
console = rich.get_console()

with handle_common_errors(console) as error_state:
    optimization_id = BespokeExecutor.submit(...)

    results = BespokeExecutor.retrieve(optimization_id)
    results = wait_until_complete(optimization_id)

if error_state["has_errored"]:
    return
```

that handles capturing and printing information about the common internal errors the executor may have in a centralised way. It also allows users using the programatic API to have more control over how the internal errors, such as a 'bespoke executor not found' error, are handled.

Finally, this PR adds a new `retrieve` command to the CLI that can save both the output and FF associated with a bespoke fit:

```
openff-bespoke executor retrieve --help
Usage: openff-bespoke executor retrieve [OPTIONS]

  Retrieve the current output of a bespoke optimization.

Options:
  --id TEXT           The id of the optimization to retrieve  [required]
  --output FILE       The JSON file to save the results to
  --force-field FILE  The path to save the force field to
  --help              Show this message and exit.
```

## Status
- [ ] Ready to go


SimonBoothroyd:
Sorry for the mega PR @jthorton ! There was quite a few things that I wanted to get done on very similar bits of code as quickly as possible, so they all kind of just ended up here.

SimonBoothroyd:
I'm going to merge this so I can finish off the other things I want to add before the release and we pass to the industry partners, but we can change anything added here down the road!

PR #120, opened 2022-02-17
Expose bespoke worker settings

SimonBoothroyd:
## Description

This PR begins to expose worker settings through the python API and CLI, including the number of cores and maximum memory available to QC engines such as `psi4`. Currently only the QC server makes use of these settings.

## Status
- [X] Ready to go


jthorton:
@SimonBoothroyd thanks for taking this on, this is a great start it would be great if we could next 
parallelize the torsiondrives so that the worker could run multiple optimisations simultaneously.

PR #134, opened 2022-03-01
Replace `settings` variable with function

### LINKED ISSUE #130: Executor run CLI not respecting `--qc-compute-n-cores`
Running the [example](https://openff-bespokefit.readthedocs.io/en/latest/getting-started/quick-start.html#one-off-fits) in the docs using bespokefit version 0.1.0 installed via conda-forge sets up 2 qc workers but gives them both all available cores rather than the 8 specified via the CLI.

SimonBoothroyd:
## Description

This PR replaces the 'dangerous' `settings` field exposed by the bespoke executor with a more reliable function that returns the most up-to-date values.

The object returned by `current_settings()` should not be modified as the changes will have no effect. Instead the `apply_env` context manager should be used e.g.

```python
with Settings(
    BEFLOW_FRAGMENTER_WORKER_N_CORES=...,
    BEFLOW_FRAGMENTER_WORKER_MAX_MEM=...
).apply_env():

    settings = current_settings()
```

Possibly fixes #130 

## Status
- [x] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/qcgenerator/worker.py
+++ b/openff/bespokefit/executor/services/qcgenerator/worker.py
@@ -21,14 +21,16 @@ from qcelemental.models.procedures import (
 from qcelemental.util import serialize
 from qcengine.config import get_global
 
-from openff.bespokefit.executor.services import settings
+from openff.bespokefit.executor.services import current_settings
 from openff.bespokefit.executor.utilities.celery import configure_celery_app
 from openff.bespokefit.schema.tasks import OptimizationTask, Torsion1DTask
 
+__settings = current_settings()
+
 redis_connection = redis.Redis(
-    host=settings.BEFLOW_REDIS_ADDRESS,
-    port=settings.BEFLOW_REDIS_PORT,
-    db=settings.BEFLOW_REDIS_DB,
+    host=__settings.BEFLOW_REDIS_ADDRESS,
+    port=__settings.BEFLOW_REDIS_PORT,
+    db=__settings.BEFLOW_REDIS_DB,
 )
 celery_app = configure_celery_app("qcgenerator", redis_connection)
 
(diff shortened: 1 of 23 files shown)


openff/bespokefit/executor/services/qcgenerator/worker.py
```diff
@@ -21,14 +21,16 @@
...
+
 redis_connection = redis.Redis(
-    host=settings.BEFLOW_REDIS_ADDRESS,
-    port=settings.BEFLOW_REDIS_PORT,
-    db=settings.BEFLOW_REDIS_DB,
+    host=__settings.BEFLOW_REDIS_ADDRESS,
+    port=__settings.BEFLOW_REDIS_PORT,
+    db=__settings.BEFLOW_REDIS_DB,
 )
 celery_app = configure_celery_app("qcgenerator", redis_connection)
```
jthorton, line 35: The info in the `redis_connection` object seems to be the same as that in the settings class and this logic seems to be repeated for each app, would it make sense to have the `configure_celery_app` just call the `current_settings` function and get the relavent info?

SimonBoothroyd: Yeah I think either having configure celery return a tuple of both the app and redis connection, or adding a new `configure_redis` method that also checks for the executor key you mentioned in slack / any other validation would be a great shout! 

I think I have a slight preference for the latter, but not by much 🙂

SimonBoothroyd: We should probably handle this in a new PR though so it doesn't get buried here! 😅

jthorton (approved):
@SimonBoothroyd thanks for the quick fix, can confirm this now works on linux as expected!

PR #137, opened 2022-03-02
Theory page in docs

### LINKED ISSUE #129: Theory section notes
The following is a list of theory section notes that should be fleshed out into a full docs section. I think the main parts of the workflow are well described in the docs already so I have mainly focused on the smirks generation and parameter cache, I have included some notes on figures that might help explain some of the ideas and will add them once we agree on which ones are needed.

Do we think we should describe the features used for each cache so users can tell when bespokefit will/won't reuse results? 

tagging @SimonBoothroyd @j-wags @Yoshanuikabundi 

## Fragmentation
- Molecules are fragmented to reduce the computational cost of generating the reference QC data.
- Currently, only torsiondrives take advantage of fragmentation.
- Bespokefit can use fragmentation engines provided by openff-fragmenter directly and by default uses the WBO fragmentation method.
- In short, it builds the minimum fragment around each non-terminal rotatable bond which is electronically decoupled from the rest of the molecule by ensuring the WBO in the bond is the same in the fragment and the parent molecule. Link back to the theory in the [fragmenter docs](https://fragmenter.readthedocs.io/en/latest/index.html#wbo-sensitive-fragmentation). 

## SMIRKS Generation
- After fragmentation we can produce the bespoke SMIRKS patterns here we focus on the torsion smirks.
- We will have one fragment per rotatable bond (note some fragments may be the same if the torsions are coupled.)
- Maybe a picture of the parent molecule with rotatable bonds highlighted and arrows from the bond to the resulting fragment. Best to pick a molecule with some repeated fragments (the TYK2 series is a good example).
- Then we use [chemper](https://github.com/MobleyLab/chemper) to automatically generate new bespoke smirks patterns for each symmetry unique group of torsions that pass through the targeted central bond. 
- First, we get the atom symmetry groups using rdkit or openeye and then assign each torsion a symmetry code in the form of a tuple of atom symmetry groups corresponding to each atom in the torsion.
- Torsions are then grouped by this torsion symmetry code and a bespoke smirks pattern is made for each symmetry unique group of torsions. 
- This helps minimise the number of parameters we introduce which helps simplify the optimisation problem while still having enough degrees of freedom to achieve good accuracy.  
- Image of a fragment with the symmetry groups labelled next to another image with some symmetrical torsions highlighted. 
- Chemper is then used to generate a pattern that targets all symmetry equivalent torsions per torsion symmetry group in both the parent and fragment using the [ClusterGraph](https://github.com/MobleyLab/chemper#clustergraph) class. 
- Layers is set to `all` to ensure the maximum common substructure between the parent and fragment is encoded into the resulting bespoke smirks pattern.
- This ensures the pattern is transferable between the fragment and parent with no further modifications but also ensures it is specific enough to make sure the pattern can only be applied to other molecules which share the electronically decoupled substructure as determined by fragmenter. 
- This is important as the parameter will account for non-local effects which can not be encoded using a standard atom type approach where by parameters are stored under the combination of the atom types that make up the torsion. 
- Image of a bespoke SMIRKS pattern and with the corresponding parent and fragment molecule. The molecules should have a substructure in the SMIRKS highlighted.
- In the case of a congeneric series like the TYK2 this means that the common core need only be parameterised once, the bespoke parameters are then cached and can be reused.
- Grid image of the TYK2 series with the matched atoms highlighted from the above SMIRKS pattern.
- By default, we take a conservative approach when generating bespoke torsion parameters and expand the k terms of our reference parameter to include all periodicities from 1 to 4 ([can be changed](https://openff-bespokefit.readthedocs.io/en/latest/ref/api/openff.bespokefit/utilities/smirks/openff.bespokefit.utilities.smirks.SMIRKSettings.html#openff.bespokefit.utilities.smirks.SMIRKSettings)).
- In many cases, this is not needed and leads to redundant parameters which are held very close to the initial guess of 0 by the L1 regulariser in forcebalance.

## QC Generation
- for each molecule/fragment (only in the case of torsiondrives) and fitting target combination a QC reference data generation task is created. 
- To avoid unnecessary torsiondrive calculations this stage will only generate tasks for fragments that can not reuse any cached bespoke parameters.
- Tasks include optimisations, hessians and 1D torsiondrives. Each task is then submitted to the QC generation queue and consumed by celery workers which run the job via [QCEngine](https://github.com/MolSSI/QCEngine) thus any qcengine supported program can be used to calculate the reference data.
- Once a task has successfully completed we also cache the result for later use which can save significant time when refitting a torsion with different optimisation settings to the same reference data. 
- For high throughput bespoke parameter deviation we recommend using the fantastic [QCFracal](https://github.com/MolSSI/QCFractal) distributed computing and database for quantum chemistry. 
- Users can easily set up a local server and submit large datasets of torsiondrives using [QCSubmit](https://github.com/openforcefield/openff-qcsubmit/) which can then be used as reference data by bespokefit via the cache update feature- need a section on how to pull down qcarchive data and cache locally. 

## Optimisation
- We use the same fitting techniques to optimise the bespoke parameters as those used to make the mainline [openff force fields](https://pubs.acs.org/doi/abs/10.1021/acs.jctc.1c00571?casa_token=kcqqp1v-6yMAAAAA:e8fNPsOOOFe4oR6kBPY4dm3HJ4jC9QDZrBhb-bVGPrHu5odnxTgXUmYZXEFjiGDt5hb5Arbyf6dPxoKB).
- In short torsion parameters are optimised by minimising the RMSE difference between the QM and MM potential energy surfaces. 
- Due to limitations in the functional form of the MM force field we perform a restrained relaxation from the QM reference minimums before computing the potential energy. 
- This helps relax high energy steric clashes while the restraints keep us close to the QM minimum. 
- By default we use quite large priors on the torsion k values so we are not hindered by our initial guess parameters which are taken from the reference force field or 0 if the term was not already present. 
- Once optimisation has been completed successfully the bespoke parameters are cached so that they can be reused on any other molecule which shares this substructure provided the optimisation settings remain the same.

Yoshanuikabundi:
## Description
Adds the theory page discussed in #129.

## Todos
  - [x] Write draft
  - [x] Iterate on draft
  - [x] Add pictures
 
## Status
- [x] Ready to go

Closes #129


Yoshanuikabundi:
@SimonBoothroyd @jthorton I think this is ready for review. I think the QC generation section is very sparse on details compared to the other sections, but I need some notes to flesh it out; I was thinking details like the number of states in the torsion drive, QC level of theory/functional/basis set etc would go nicely there.

SimonBoothroyd:
Brilliant thanks @Yoshanuikabundi ! I'll take a look in the next couple of days.

Yoshanuikabundi:
OK, I've (finally) got around to final edits, I think this should be ready to merge!

jthorton (approved):
@Yoshanuikabundi great job feel free to merge when ready!

PR #138, opened 2022-03-07
Add CLI override for default bespoke QC spec

SimonBoothroyd:
## Description

This PR adds a new `--default-qc-spec` CLI flag that easily allows users to override the default QC spec set by a workflow / workflow file without needing to create a new one in python.

This should make it easier for less-python centric users to experiment with different QC methods.

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/cli/executor/submit.py
+++ b/openff/bespokefit/cli/executor/submit.py
@@ -75,6 +75,15 @@ def submit_options():
             required=False,
             multiple=True,
         ),
+        optgroup.option(
+            "--default-qc-spec",
+            type=(str, str, str),
+            help="The program, method, and basis to use by default when performing any "
+            "QC calculations, e.g. `--default-qc-spec xtb gfn2xtb none`. If no basis "
+            "is required to be specified for a particular method (e.g. for ANI or XTB) "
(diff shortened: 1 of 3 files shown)


openff/bespokefit/cli/executor/submit.py
```diff
@@ -119,6 +131,21 @@ def _to_input_schema(
...
             workflow_factory.target_torsion_smirks = [*target_torsion_smirks]
+        if default_qc_spec is not None:
+
+            program, method, basis = default_qc_spec
+
+            if basis.lower() == "none":
+                basis = None
+
+            workflow_factory.default_qc_specs = [
+                QCSpec(
```
jthorton, line 142: if the user enters an invalid combination of program, method and basis the spec will raise an `QCSpecificationError ` [here](https://github.com/openforcefield/openff-qcsubmit/blob/439b117a79e554b4595d5871a6eeca87f123cbd3/openff/qcsubmit/common_structures.py#L496) which should also be caught and displayed as it will give hints on valid combinations. We probably also want a test for this.

jthorton (changes requested):
Awsome this will really help the CLI users, I think once the exception catch is added this will be good to go.

PR #140, opened 2022-03-09
Multi-molecule submissions via the CLI

jthorton:
## Description
This PR extends the `openff-bespoke executor submit` command to accept and process multiple molecules at once via multiple smiles or input file arguments. Input files may also contain multiple molecules. A formatted table is printed to the terminal and optionally written to a CSV file, which maps the input molecule to an optimization ID and input file if appropriate. 

image

## Status
- [x] Ready to go

### DIFF
--- a/docs/getting-started/quick-start.md
+++ b/docs/getting-started/quick-start.md
@@ -126,8 +126,12 @@ openff-bespoke executor submit --file        "acetaminophen.sdf"   \
                                --workflow    "default"
 ```
 
-The `submit` command will print a unique ID that has been assigned by the executor to the submission. This ID can be 
-used to check on state of the submission:
+The `submit` command will also accept a combination of the two input forms as well as multiple occurrences of either. 
+After successfully submitting the molecules a table will be printed which maps the unique ID that has been assigned by 
+the executor to the submission to the molecule smiles and input file if appropriate. To keep track of submissions we can 
+also have the table saved to csv by add the corresponding `--save-submission-info` flag to the command.
+
+The ID's can be used to check on state of the submission:
 
 ```shell
 openff-bespoke executor watch --id "1"
--- a/openff/bespokefit/cli/executor/submit.py
+++ b/openff/bespokefit/cli/executor/submit.py
@@ -176,21 +176,24 @@ def _submit(
 
     console.print(Padding("1. preparing the bespoke workflow", (0, 0, 1, 0)))
 
+    all_molecules = []
+
     if input_file_path:
         with console.status("loading the molecules"):
-            all_molecules = []
             for input_file in input_file_path:
                 file_molecules = Molecule.from_file(input_file)
                 if isinstance(file_molecules, Molecule):
-                    file_molecules.properties["input_file"] = input_file
-                    all_molecules.append(file_molecules)
-                else:
-                    for m in file_molecules:
-                        m.properties["input_file"] = input_file
-                    all_molecules.extend(file_molecules)
-    else:
+                    file_molecules = [file_molecules]
+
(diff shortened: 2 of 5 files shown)


openff/bespokefit/cli/executor/submit.py
```diff
@@ -156,67 +160,135 @@ def _to_input_schema(
...
+        f"[[green]✓[/green]] [blue]{len(all_molecules)}[/blue] molecules found"
+    )
 
-    with console.status("building fitting schemas"):
+    input_schemas = []
+    for molecule in track(
+        all_molecules,
+        description="building fitting schemas",
+        console=console,
+        total=len(all_molecules),
```
SimonBoothroyd, line 221:
```suggestion
        total=len(all_molecules),
        transient=True,
```

It might be nice to hide the progress bar after it gets to the end as we show a green tick on the next line, and I think(?) adding `transient=True` should do this?

openff/bespokefit/cli/executor/submit.py
```diff
@@ -156,67 +160,135 @@ def _to_input_schema(
...
-    console.print("[[green]✓[/green]] fitting schema generated")
+    console.print("[[green]✓[/green]] fitting schemas generated")
 
     console.print(Padding("2. submitting the workflow", (1, 0, 1, 0)))
+    response_ids = []
+    for input_schema in track(
+        input_schemas,
+        description="submitting tasks",
+        total=len(input_schemas),
+        console=console,
```
SimonBoothroyd, line 243:
```suggestion
        console=console,
        transient=True,
```

openff/bespokefit/cli/executor/submit.py
```diff
@@ -225,8 +297,8 @@ def _submit_cli(
     console = rich.get_console()
     print_header(console)
 
-    if (input_file_path is not None and molecule_smiles is not None) or (
-        input_file_path is None and molecule_smiles is None
+    if (input_file_path and molecule_smiles) or (
+        not input_file_path and not molecule_smiles
     ):
```
SimonBoothroyd, line 302: Hmm I wonder if we can just drop this check + error now?

SimonBoothroyd (approved):
Looks like a great addition :tada:! If you have a moment could you add a sentence or two to the docs (Quick Start + Executor pages maybe?) just to describe that this is now possible?

openff/bespokefit/cli/executor/submit.py
jthorton: Yeah, I think it would be safe to allow both input types.

openff/bespokefit/cli/executor/submit.py
jthorton: Nice that would look good Ill try it out!

docs/getting-started/quick-start.md
```diff
@@ -126,8 +126,12 @@ openff-bespoke executor submit --file        "acetaminophen.sdf"   \
...
 ```
 
-The `submit` command will print a unique ID that has been assigned by the executor to the submission. This ID can be 
-used to check on state of the submission:
+The `submit` command will also accept a combination of the two input forms as well as multiple occurrences of either. 
+After successfully submitting the molecules a table will be printed which maps the unique ID that has been assigned by 
+the executor to the submission to the molecule smiles and input file if appropriate. To keep track of submissions we can 
+also have the table saved to csv by add the corresponding `--save-submission-info` flag to the command.
+
+The ID's can be used to check on state of the submission:
```
SimonBoothroyd, line 134: 🎉 thanks! This should make it easier for new users to discover 😁

PR #142, opened 2022-03-10
Deduplicate symmetry equivalent fragments

### LINKED ISSUE #141: Repeated calculations for symmetry equivalent fragments
When passing in a molecule that has symmetry and symmetry equivalent fragments bespokefit is unable to deduplicate these tasks resulting in repeated calculations. We also generate redundant SMIRKS patterns for each symmetry equivalent fragment meaning that only the last pattern in the hierarchy will actually match and be optimized. This is also an issue in the optimization stage as we take the gradient of parameters that are not actually present in the system, this is wasteful and avoiding this would speed up the optimization stage. An example of this is Dimethyl hydrogen phosphate which runs two torsiondrives with default fragmentation settings when only one is needed.

I think that by simply removing symmetry equivalent fragments from the fragmentation result we should avoid this issue as the parameters derived for one fragment will still transfer to the rest.

jthorton:
## Description
This PR fixes #141 by filtering fragments from the fragmentation stage which are symmetry equivalent. This should stop duplicate torsiondrives being run, as they can make it through the hash check as the tagged atoms are different, and stop duplicated parameters from being made and fit.  

## Status
- [ ] Ready to go


openff/bespokefit/executor/services/fragmenter/worker.py
```diff
@@ -67,8 +69,49 @@ def fragment(
...
+        ].append(fragment_result)
+
+    unique_fragments = []
+    # remove duplicated symmetry equivalent fragments
+    for fragments in fragments_by_smiles.values():
+        if len(fragments) == 1:
+            unique_fragments.extend(fragments)
+
+        else:
+            symmetry_groups = set()
```
SimonBoothroyd, line 99:
```suggestion
        if len(fragments) == 1:
            unique_fragments.extend(fragments)
            continue

        symmetry_groups = set()
```
As a very pedantic thing you could probably add a continue to the if statement and then remove / de-indent the else statement!

openff/bespokefit/tests/executor/services/fragmenter/test_worker.py
```diff
@@ -65,3 +71,42 @@ def test_fragmentation_skip(bace):
...
+
+def test_fragmentation_equivalent_no_symmetry():
+    """Make sure duplicated fragments which are not symmetry equivalent are not filtered."""
+
+    molecule = Molecule.from_file(
+        file_path=get_data_file_path(
+            os.path.join("test", "molecules", "tyk2", "ejm31.sdf"), "openff.bespokefit"
+        ),
+        file_format="sdf",
+    )
```
SimonBoothroyd, line 101: Is there perhaps a smaller / cheaper molecule which should test this behaviour so we don't have to fragment a larger molecule everytime and keep the unit tests as quick as possible? +1 if can also add a check that the output de-duplicated fragments match a set of expected ones!

SimonBoothroyd (approved):
LGTM! Just a few minor suggestions you may want to consider before merging 🙂

openff/bespokefit/tests/executor/services/fragmenter/test_worker.py
jthorton: Good spot yeah I think that is what the bace files were originally for I have updated the test to use these which is a slightly smaller molecule and still covers what I wanted to test!

Issue #128, opened 2022-02-22
Parallelize torsiondrives


jthorton:
@SimonBoothroyd  made a great start exposing QC worker settings including memory and number of cores for each job #120. It would be good if we could also parallelize the torsiondrives so each worker could run multiple optimisations simultaneously. This might mean we need to rework the qcengine procedure slightly to achieve this possibly by subclassing the current one and registering a new parallel implementation under a different alias. We may also want to introduce a new setting like `N_QC_PARALLEL_TASKS` which determines the maximum number of tasks to run in parallel and how the maximum resources should be divied.

SimonBoothroyd:
If we can refactor [these lines](https://github.com/MolSSI/QCEngine/blob/aeead6fa1327917c7b9fd874479e013114b45b3e/qcengine/procedures/torsiondrive.py#L70-L73) into a new function, e.g.

```python
def _spawn_optimizations(input_model, config, next_jobs):
    grid_point_results = {
        grid_point: [self._spawn_optimization(grid_point, job, input_model, config) for job in jobs]
        for grid_point, jobs in next_jobs.items()
    }
    return grid_point_results
```

then we could easily parallelize the procedure in a sub-class by overidding like:

```python
def _spawn_optimizations(input_model, config, next_jobs):

    with Pool(processes=n_processes) as pool: 

        grid_point_results = {
            grid_point: list(pool.map(self._spawn_optimization, jobs))
            for grid_point, jobs in next_jobs.items()
        }

    return grid_point_results
```

jthorton:
Awesome that's just what I was thinking, just testing locally now, another thing we can have is that the pool size scales with how many jobs there are to run each iteration of the torsiondrive. So in step one for example we might have 2 tasks but a max pool size of 4 jobs given by the user so we would set the pool to 2 and divide the resources equally to ensure we maximise their usage.

PR #143, opened 2022-03-14
Parallel torsiondrives

jthorton:
## Description
This PR implements #128 by overriding the new `_spawn_optimizations` function in the torsiondrive QCEngine procedure (version 0.23.0 or later). A new setting is added `BEFLOW_QC_COMPUTE_WORKER_N_TASKS` which controls the maximum number of parallel optimizations each worker can run, where the total resources given to a worker via the variables `BEFLOW_QC_COMPUTE_WORKER_N_CORES` & `BEFLOW_QC_COMPUTE_WORKER_MAX_MEM` are equally divided across each optimization.

Note this speed up is only useful for QM methods where the cost of the optimization is higher than the overhead of creating a multiprocess pool. 

## Status
- [x] Ready to go


openff/bespokefit/executor/services/qcgenerator/qcengine.py
```diff
@@ -0,0 +1,83 @@
...
+        next_jobs: Dict[str, List[float]],
+        input_model: TorsionDriveInput,
+        config: TaskConfig,
+    ) -> Dict[str, List[Union[FailedOperation, OptimizationResult]]]:
+        """
+        Spawn parallel optimizations based on the number of next jobs and available workers.
+        """
+
+        _settings = current_settings()
+        opts_per_worker = _settings.BEFLOW_QC_COMPUTE_WORKER_N_TASKS
```
SimonBoothroyd, line 45:
```suggestion
        settings = current_settings()
        opts_per_worker = settings.BEFLOW_QC_COMPUTE_WORKER_N_TASKS
```

openff/bespokefit/executor/services/qcgenerator/qcengine.py
```diff
@@ -0,0 +1,83 @@
+from collections import defaultdict
```
SimonBoothroyd, line 1:
```suggestion
```

openff/bespokefit/executor/services/qcgenerator/qcengine.py
```diff
@@ -0,0 +1,83 @@
...
+
+        _settings = current_settings()
+        opts_per_worker = _settings.BEFLOW_QC_COMPUTE_WORKER_N_TASKS
+        final_results = defaultdict(list)
+        n_jobs = sum([len(value) for value in next_jobs.values()])
+        if opts_per_worker > 1 and n_jobs > 1:
+            # split the resources based on the number of tasks
+            n_workers = int(min([n_jobs, opts_per_worker]))
+            opt_config = _divide_config(config=config, n_workers=n_workers)
+            work_list = []
```
SimonBoothroyd, line 52:
```suggestion
```

openff/bespokefit/executor/services/qcgenerator/qcengine.py
```diff
@@ -0,0 +1,83 @@
...
+
+                # now loop over the work and pull results and replace values
+                for work in work_list:
+                    result = work.get()
+                    for jobs in final_results.values():
+                        if work._job in jobs:
+                            index = jobs.index(work._job)
+                            jobs[index] = result
+
+            return final_results
```
SimonBoothroyd, line 71: I think we could simplify this a little and avoid accessing `async_result._job` :

```suggestion
                tasks = {
                    grid_point: [
                        pool.apply_async(
                            func=self._spawn_optimization,
                            args=(grid_point, job, input_model, opt_config),
                        )
                        for job in jobs
                    ]
                    for grid_point, jobs in next_jobs.items()
                }
                return {
                    grid_point: [grid_task.get() for grid_task in grid_tasks]
                    for grid_point, grid_tasks in tasks.items()
                }
```

openff/bespokefit/executor/services/qcgenerator/qcengine.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Override the _spawn_optimizations method of the basic torsiondrive procedure to allow for parallel optimizations
+    within one worker.
+    """
+
+    _defaults = {"name": "TorsionDriveParallel", "procedure": "torsiondrive"}
+
+    def _spawn_optimizations(
+        self,
+        next_jobs: Dict[str, List[float]],
```
SimonBoothroyd, line 36:
```suggestion
        next_jobs: Dict[str, List[List[float]]],
```

openff/bespokefit/executor/services/_settings.py
```diff
@@ -54,6 +54,7 @@ class Settings(BaseSettings):
     BEFLOW_QC_COMPUTE_WORKER = "openff.bespokefit.executor.services.qcgenerator.worker"
     BEFLOW_QC_COMPUTE_WORKER_N_CORES: Union[int, Literal["auto"]] = "auto"
     BEFLOW_QC_COMPUTE_WORKER_MAX_MEM: Union[float, Literal["auto"]] = "auto"
+    BEFLOW_QC_COMPUTE_WORKER_N_TASKS: float = 1
```
SimonBoothroyd, line 57: Maybe for now we should let the spawn method decide this by default? e.g. if the program is one of xtb, ani, or OpenFF for example we should maybe have this be equal to the number of cores, whereas if the program is psi4 this should be 1?

```suggestion
    BEFLOW_QC_COMPUTE_WORKER_N_TASKS: Literal["auto"] = "auto"
```

SimonBoothroyd (commented):
This is looking great, thanks for adding this! Just a couple of minor suggestions and one question about determining n_tasks automatically that may be good to discuss before merging!

openff/bespokefit/executor/services/_settings.py
jthorton: Yes, that could work well, I think we will definitely want to use this with psi4 as well though as the speedup should be significant when the optimization cost is much larger than the time to make the pool. For small molecules splitting the optimisations can slow down the torsiondrive with quick methods like xtb. Also if we go down the auto route we could not expose the variable in the first pass?

SimonBoothroyd: > psi4 as well though as the speedup should be significant when the optimization cost is much larger than the time to make the pool

Ah good point - maybe something like `n_tasks = n_cores / 8` when using `psi4` as that's how many we normally give workers on QCF?

> For small molecules splitting the optimisations can slow down the torsiondrive with quick methods like xtb. Also if we go down the auto route we could not expose the variable in the first pass?

Oh interesting, that's good to know!

>  Also if we go down the auto route we could not expose the variable in the first pass?

Hmm I think it's still handy to expose as it means we can tweak this easily when trying to optimize different QC methods? Sorry I meant to suggest changing this to `Union[int, Literal["auto"]] = "auto"`!

openff/bespokefit/executor/services/qcgenerator/qcengine.py
jthorton: Awesome that's much better than using `_job`!

openff/bespokefit/executor/services/_settings.py
jthorton: >  Ah good point - maybe something like n_tasks = n_cores / 8 when using psi4 as that's how many we normally give workers on QCF?


Yeah, I think that would be a good idea maybe more generally we could change the setting to control the number of cores per optimization then the number of parallel jobs can be determined from that?

PR #144, opened 2022-03-16
Overhaul executor <-> redis interface

SimonBoothroyd:
## Description

This PR:

* abstracts the coordinator <-> redis interface into own its module
* fixes a gateway hanging error when processing many tasks due to async congestion
* adds safety checks to ensure version compatibility
* adds user-friendly messages when possible redis issues are encountered

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/_settings.py
+++ b/openff/bespokefit/executor/services/_settings.py
@@ -28,7 +28,7 @@ class Settings(BaseSettings):
     BEFLOW_GATEWAY_LOG_LEVEL: str = "error"
 
     BEFLOW_REDIS_ADDRESS: str = "localhost"
-    BEFLOW_REDIS_PORT: int = 6379
+    BEFLOW_REDIS_PORT: int = 6363
     BEFLOW_REDIS_DB: int = 0
 
     BEFLOW_COORDINATOR_PREFIX = "tasks"
@@ -38,6 +38,8 @@ class Settings(BaseSettings):
     BEFLOW_COORDINATOR_WORKER = (
         "openff.bespokefit.executor.services.coordinator.worker:"
     )
+    BEFLOW_COORDINATOR_MAX_UPDATE_INTERVAL: float = 5.0
+    BEFLOW_COORDINATOR_MAX_RUNNING_TASKS: int = 1000
 
     BEFLOW_FRAGMENTER_PREFIX = "fragmentations"
     BEFLOW_FRAGMENTER_ROUTER = (
--- a/openff/bespokefit/executor/services/coordinator/worker.py
+++ b/openff/bespokefit/executor/services/coordinator/worker.py
@@ -1,77 +1,116 @@
 import asyncio
 import logging
-import pickle
+import time
 
 import redis
 
 from openff.bespokefit.executor.services import current_settings
-from openff.bespokefit.executor.services.coordinator.models import CoordinatorTask
+from openff.bespokefit.executor.services.coordinator.storage import (
+    TaskStatus,
+    get_n_tasks,
+    get_task,
+    peek_task_status,
+    pop_task_status,
+    push_task_status,
+    save_task,
+)
 
 _logger = logging.getLogger(__name__)
 
-__settings = current_settings()
 
-redis_connection = redis.Redis(
-    host=__settings.BEFLOW_REDIS_ADDRESS,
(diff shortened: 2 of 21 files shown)


openff/bespokefit/executor/services/_settings.py
```diff
@@ -28,7 +28,7 @@ class Settings(BaseSettings):
     BEFLOW_GATEWAY_LOG_LEVEL: str = "error"
 
     BEFLOW_REDIS_ADDRESS: str = "localhost"
-    BEFLOW_REDIS_PORT: int = 6379
+    BEFLOW_REDIS_PORT: int = 6363
```
SimonBoothroyd, line 31: By changing the default port hopefully we're less likely to accidentally connect to someones running redis server unintentionally!

jthorton: Great idea, this should be much safer.

openff/bespokefit/executor/services/_settings.py
```diff
@@ -38,6 +38,8 @@ class Settings(BaseSettings):
     BEFLOW_COORDINATOR_WORKER = (
         "openff.bespokefit.executor.services.coordinator.worker:"
     )
+    BEFLOW_COORDINATOR_MAX_UPDATE_INTERVAL: float = 5.0
```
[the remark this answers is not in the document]
SimonBoothroyd: Great question! We should be safe here now that we have [this line added](https://github.com/openforcefield/openff-bespokefit/blob/990238fc089d081c6fd1ddcea5a9642b45d09042/openff/bespokefit/executor/services/coordinator/worker.py#L87), this mainly just lets users tweak how much CPU the coordinator will use on average when only a few molecules are being fit!

openff/bespokefit/executor/services/coordinator/worker.py
```diff
@@ -1,78 +1,117 @@
...
+
+            n_running_tasks = get_n_tasks(TaskStatus.running)
+            n_tasks_to_queue = min(
+                settings.BEFLOW_COORDINATOR_MAX_RUNNING_TASKS - n_running_tasks,
+                get_n_tasks(TaskStatus.waiting),
+            )
+
+            for _ in range(n_tasks_to_queue):
+
+                push_task_status(
```
jthorton, line 106: This is fantastic, great idea to use separate queues!

openff/bespokefit/executor/utilities/redis.py
```diff
@@ -5,12 +5,87 @@
...
+
+
+def connect_to_redis(
+    host: str, port: int, db: int, validate: bool = True
+) -> redis.Redis:
+    """Connects to a redis server using the specified settings."""
+
+    connection_key = (host, port, db, validate)
+
+    if connection_key in __CONNECTION_POOL:
```
jthorton, line 56: Is there any danger with multiple workers using the same connection object to redis?

openff/bespokefit/executor/utilities/redis.py
```diff
@@ -93,4 +172,16 @@ def launch_redis(
...
+    try:
+        connect_to_redis("localhost", port, 0, validate=True)
+    except RedisNotConfiguredError:
+
+        if redis_save_exists:
+            raise
+
+        connection = connect_to_redis("localhost", port, 0, validate=False)
+        connection.set(
+            "openff-bespokefit:redis-version", expected_redis_config_version()
```
jthorton, line 184: Great idea to stamp a version on the database as well!

jthorton:
This looks great, the new coordinator storage file looks like the start of a client object that we could add later if there is interest! I also think it might be worth extending the `executor list` command to take a status so users can check on specific queues or all tasks if they pass `all` as listing every task for large databases may take some time?

openff/bespokefit/executor/utilities/redis.py
SimonBoothroyd: I'm not 100% certain, but I'm pretty sure it won't affect anything as the connection shouldn't really be doing anything other than pinging the redis server!

SimonBoothroyd:
> I also think it might be worth extending the executor list command to take a status so users can check on specific queues or all tasks if they pass all as listing every task for large databases may take some time?

Definitely, I can open a separate PR for this once this is merged so this one doesn't get even bigger!

PR #145, opened 2022-03-17
Add option to `list` CLI to filter by status

SimonBoothroyd:
## Description

This PR allows users to more easily filter which optimisations currently have what status when using the `list` executor command.

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/cli/executor/list.py
+++ b/openff/bespokefit/cli/executor/list.py
@@ -61,6 +61,10 @@ def list_cli(status_filter: Status):
 
     settings = current_settings()
 
+    # In the coordinator we keep both successful and errored tasks in the same 'complete'
+    # queue to avoid having to maintain and query to separate lists in redis, so here we
+    # need to condense these two states into one and then apply a second filter when
+    # iterating over the returned ids.
     status_url = (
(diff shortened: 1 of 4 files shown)


openff/bespokefit/cli/executor/list.py
```diff
@@ -49,11 +61,19 @@ def list_cli():
 
     settings = current_settings()
 
+    status_url = (
+        None
+        if status_filter is None
+        else status_filter.replace("success", "complete").replace("errored", "complete")
```
jthorton, line 67: Just to make sure I understand, we keep errored tasks in the complete queue so we have to query that queue and then filter if the actual status of the task is errored?

jthorton: slightly picky can I add a comment saying this as I am sure I will forget in future!

jthorton (approved):
Awesome change which will help find errors even faster!

openff/bespokefit/cli/executor/list.py
SimonBoothroyd: Absolutey! I've added one in fabcedc

PR #161, opened 2022-04-22
Create single-file installers

mattwthompson:


mattwthompson:
The last run successfully built an installer and installed against it, but I haven't yet got tests running and passing. I'm assuming no OpenEye license is needed to run tests.

mattwthompson:
I have it all the way to some tests passing, but some are still failing. Could you have a look at the failing tests and see what I'm missing @jthorton? I can add the OpenEye license in if it's required, but I don't know if that would cover all test failures.

The other option would be to not run tests based on the installer - I'd prefer keeping them in at all possible.

jthorton:
@mattwthompson Thanks for putting this together, and sure happy to help but not sure where I should be looking for tests? in the single file installer CI I see it failing before getting to the tests or do I need to download the installer and run the tests locally?

mattwthompson:
Updated https://github.com/openforcefield/toolkit-installer-constructor/commit/5e072dfbd7c7e203c1e137ab28d5e551bcb8cabe and restarting now

mattwthompson:
Most recent failures - PDB loading, QCArchive offline https://gist.github.com/mattwthompson/f39e6bc4a89275a2f4339e6578c9460b

mattwthompson:
Some tests are failing because they require OpenEye behavior - I'll open a PR to skip these tests in that case.

mattwthompson:
The single-file installers are being built (with AmberTools installed - no OpenEye license or software packaged) and seem to not only exist but also pass (most) tests). Most of the failures seem to be a hard OpenEye dependency, which I attempted to fix in #170. There's one test failing in that PR but otherwise is good, I think.

mattwthompson:
Okay, #170 is merged in and hopefully the tests pass here now.

mattwthompson:
Tests are passing; I'll do some final touch-ups, merge, and upload the installers to the latest release. As this does not touch functionality of the software itself, I don't think waiting for reviews is warranted, though feedback is always the case on a non-draft PR.

Please note that these are XTB + AmberTools (+ RDKit) bundles; adding a Psi4 bundle to this array is a separate task and, as noted above, none will be made with OpenEye Toolkits bundled in. I've updated the file names to reflect this, using the pattern

```
name: ${{ matrix.os }}_py${{ matrix.python-version }}_xtb.sh

File: .github/workflows/installer.yml [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+        if [[ $LATEST_TAG != $FOUND_VER ]];
+          then echo "Version mismatch"
+          echo $LATEST_TAG
+          echo $FOUND_VER
+          exit 1
+        fi
+
+        # The toolkit's action does not require installing optional and test-only
+        # dependencies, not sure if/why they need to be brought in here
+        conda install "forcebalance >=1.9.2" "openff-fragmenter-base >=0.1.2" -c conda-forge -yq
```
j-wags, line 87: Shouldn't these already be present? See package deps in https://anaconda.org/conda-forge/openff-bespokefit/files

j-wags (approved):
Fantastic. Thanks a million, @mattwthompson!!

.github/workflows/installer.yml
mattwthompson: You're right; I checked the logs and this step did nothing ("all packages are already installed" or whatever the message is). This line probably stems from something being botched in earlier iterations

Issue #166, opened 2022-04-26
Parallelisation with --n-qc-compute-workers


xiki-tempula:
I'm running the [openff-bespokefit](https://github.com/openforcefield/openff-bespokefit) with 12 works on a 96 core machine with the command
`openff-bespoke executor run --file "lig.sdf" --workflow "default" --output "lig.json" --output-force-field "lig.offxml" --n-qc-compute-workers 12 --qc-compute-n-cores 8 --default-qc-spec psi4 B3LYP-D3BJ DZVP`
I noticed that only 5 workers are being used (5 psi4 processes were being used). I wonder how is the parallelisation being set up, so I could better allocate the resources? Thank you.

jthorton:
Hi @xiki-tempula good question! this is a tricky one and its hard to know in advance how to best split the resources as this depends on the number of torsiondrive tasks produced for the molecule. Currently, each worker can consume 1 torsiondrive task at a time so the fact you have 5 activate tasks probably means the molecule makes 5 torsiondrives, so it would be better to decrease the number of workers but give them each more cores. 

We can also add some parallelisation to the torsiondrive tasks by performing multiple constrained optimisations simultaneously by editing an environment variable (note there are a lot of variables in bespokefit see [here](https://openff-bespokefit--156.org.readthedocs.build/en/156/users/bespoke-executor.html#configuring-from-the-environment)) the important one is `BEFLOW_QC_COMPUTE_WORKER_N_TASKS` which controls how many parallel optimisations each worker can do in a torsiondrive. So doing `export BEFLOW_QC_COMPUTE_WORKER_N_TASKS=2` before a run (you can also set this in your bashrc) would allow each worker to run up to 2 optimisations at a time.

I would also look at adding this to your run command [--qc-compute-max-mem](https://openff-bespokefit--156.org.readthedocs.build/en/156/getting-started/bespoke-cli.html#cmdoption-openff-bespoke-executor-run-qc-compute-max-mem) this controls how much memory per-core workers can use, by default it will try and give every worker access to all of the memory which can lead to segfaults.

xiki-tempula:
@jthorton Thanks for the explanation. I have been using TorsionDrive during my research and have done some dihedral parameterisation myself.
The TorsionDrive uses Work Queue from cctools for parameterisation. If I do
`work_queue_worker --cores=96` to set up the worker
[...]

PR #168, opened 2022-04-29
Fix parallel torsiondrives

jthorton:
## Description
This fixes a bug where the custom parallel torsiondrive code was not used. We also have to change the qc workers to be non daemon to allow them to use a process pool to run parallel optimisations and use the pool='solo' mode to launch the workers which means they run in the main task. See [here](https://www.distributedpython.com/2018/10/26/celery-execution-pool/) for more details. This does mean that the qc workers could be left as zombie processes and might not shut down correctly. In testing the workers also complain about missing heartbeats and the clocks going out of sync this should be fine as we don't use timers or countdowns. The workers also send a message to `sys.__stderr__` after they finish each task which floods the terminal. We can stop this by setting `sys.__stderr__=None` but this will also hide normal shutdown messages at the end of a run. 

Related to #166 

## Todos
  - [ ] TODO 1

## Questions
- [ ] How should we handle the stderr messages? 
- [ ] Is this the best way around the daemon lock?

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/executor.py
+++ b/openff/bespokefit/executor/executor.py
@@ -299,9 +299,19 @@ class BespokeExecutor:
                     worker_app, celery.Celery
                 ), "workers must be celery based"
 
-                self._worker_processes.append(
-                    spawn_worker(worker_app, concurrency=n_workers)
-                )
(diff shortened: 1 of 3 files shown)


jthorton:
Try threadpool worker?

PR #169, opened 2022-04-30
Fix SMIRNOFF angle prior literal validator

SimonBoothroyd:
## Description

This PR fixes a copy paste typo on `AngleHyperparameters` whereby the prior currently expects 'length' rather than 'angle'

## Status
- [X] Ready to go


jthorton:
Good catch!

PR #170, opened 2022-05-13
Include non-OpenEye tests in CI

mattwthompson:


mattwthompson:
I could use some help with this failure: https://gist.github.com/mattwthompson/ae91071425995e0994594f471dd062e6

jthorton:
@mattwthompson I think this is probably related to the reference hash in the test being generated in part using OpenEye, I could generate two hashes nd check for either then it should work with rdkit or openeye

mattwthompson:
@jthorton I'll take you up on that - it would take me much longer to figure out how to generate that hash than asking you to do it. Feel free to commit directly to this branch!

jthorton:
@mattwthompson this should be good to go now, hopefully the hash won't need to be regenerated each time a toolkit changes version!

PR #172, opened 2022-05-16
[WIP] Update quick start guide to use psi4-ambertools stack

j-wags:
## Description
Provide a brief description of the PR's purpose here.

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -6,13 +6,38 @@ package manager.
 
 ## Using conda
 
-The recommended way to install `openff-bespokefit` is via the `conda` package manager:
+The recommended way to install `openff-bespokefit` is via the `conda` package manager.
+A working installation also requires at least one package from each of the two sections below 
+("Fragmentation Backends" and "Reference Data Generators")
 
 ```shell
 conda install -c conda-forge openff-bespokefit
 ```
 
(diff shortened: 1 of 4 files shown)


mattwthompson (commented):
I was able to install everything (on my second try) and get the "one-off" example and Python API sections to run. I took notes while working through the installation process and quick start guide. They are [here](https://gist.github.com/mattwthompson/fa6a7bd7c474b879c5a94c505103290d), unedited, but the relevant feedback I have to give is:

* The installation guide gave rough benchmark for AmberTools fragmentation but none for OpenEye. This is minor but when reading it I was expecting a number there as well.
* On startup, the CLI printed (`OpenFF Bespoke`) - this probably should be something like `OpenFF BespokeFit` for consistency?
* It would be nice to have a copy-paste button on the code sections the user is meant to copy. I'm referring to this button elsewhere:
image
* For the commands that fit one molecule, it would be nice to have an idea of how long each step might take. I have no familiarity with XTB and have no clue if 10+ minute runtime is normal or if I screwed up something during installation. (As a grad student, there's a decent chance I would have killed the task and moved on to something else if a part of a "quick start" guide took this long.)

PR #174, opened 2022-05-18
Use mamba in CI

mattwthompson:
In #161 I noticed that same actions took a long time (~35 minutes) to prepare the conda env, probably because they don't use `mamba` by default. ([Or so I think.](https://github.com/conda-incubator/setup-miniconda#example-6-mamba)). These test environments are fairly complex but I'm still surprised they took so long.

I didn't find anything in closed issues about this - apologies if this has already been attempted.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -34,6 +34,7 @@ jobs:
         if: ${{ matrix.openeye == true }}
         with:
           python-version: ${{ matrix.python-version }}
+          mamba-version: "*"
           environment-file: devtools/conda-envs/test-env.yaml
 
           channels: openeye,conda-forge,defaults
--- a/.github/workflows/installer.yml
+++ b/.github/workflows/installer.yml
@@ -30,6 +30,7 @@ jobs:
     - uses: conda-incubator/setup-miniconda@v2.1.1
       with:
         python-version: ${{ matrix.python-version }}
+        mamba-version: "*"
         activate-environment: constructor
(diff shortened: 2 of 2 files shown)


jthorton:
Thanks, @mattwthompson I didn't notice we weren't using it so this is defiantly worth adding, seems like the linux env with no openeye still take around 30 mins to complete though I wonder if we could make that one faster somehow

mattwthompson:
Nice catch:

```
2022-05-18T04:10:05.4346814Z ##[group]Updating 'test' env from conda env update...
2022-05-18T04:10:05.4348526Z [command]/usr/share/miniconda/condabin/conda env update --name test --file /home/runner/work/openff-bespokefit/openff-bespokefit/devtools/conda-envs/setup-miniconda-patched-no_openeye.yaml
2022-05-18T04:10:57.4405297Z Collecting package metadata (repodata.json): ...working... done
2022-05-18T04:34:52.3883729Z Solving environment: ...working... done
```

I'll see if I can figure something out.

mattwthompson:
Cheating a bit and adding some lower constraints (3edfb82b034df8b2b48d3044f6676c11b763e47d) helped a bit, dropping it down to 15 minutes.

I noticed in a local environment creation that it was fast/instant, but also didn't install anything from the `defaults` channel. So I dropped it (e2ea98138afaa8e59595d0bd339a5d561516a36a) and the step dropped again to 10 minutes.

Let's see if that affects the other environments?

PR #167, opened 2022-04-27
Add single point jobs to the end of torsiondrives

jthorton:
## Description
This PR expands the torsiondrive task with an optional second QC specification which is used to perform single point energy evaluations on the optimised geometries to resolve the PES.

## Todos
Notable points that this PR has either accomplished or will accomplish.
  - [ ] add tests
  - [ ] fix the CLI arguments to reflect the new options

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

PR #176, opened 2022-05-19
Split torsion drive single-points into separate tasks

SimonBoothroyd:
## Description

This PR alters PR #167 such that torsion drive optimisation and single point evaluation caching is handled by celery directly.

## Status
- [X] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/qcgenerator/cache.py
+++ b/openff/bespokefit/executor/services/qcgenerator/cache.py
@@ -5,7 +5,12 @@ import redis
 from openff.toolkit.topology import Molecule
 
 from openff.bespokefit.executor.services.qcgenerator import worker
-from openff.bespokefit.schema.tasks import HessianTask, OptimizationTask, Torsion1DTask
+from openff.bespokefit.schema.tasks import (
+    HessianTask,
+    OptimizationTask,
+    QCGenerationTask,
+    Torsion1DTask,
+)
 from openff.bespokefit.utilities.molecule import canonical_order_atoms
 
 if TYPE_CHECKING:
@@ -54,8 +59,9 @@ def _canonicalize_task(task: _T) -> _T:
     return task
 
 
-def _sha256_hash(contents: str) -> str:
-    return hashlib.sha512(contents.encode()).hexdigest()
+def _hash_task(task: QCGenerationTask) -> str:
+    """Returns a hashed representation of a QC task"""
+    return hashlib.sha512(task.json().encode()).hexdigest()
 
 
 def _retrieve_cached_task_id(
@@ -71,6 +77,7 @@ def _retrieve_cached_task_id(
 def _cache_task_id(
     task_id: str, task_type: str, task_hash: str, redis_connection: redis.Redis
 ):
+    """Store the ID of a running QC task in the QC task cache."""
 
     redis_connection.hset("qcgenerator:types", task_id, task_type)
     # Make sure to only set the hash after the type is set in case the connection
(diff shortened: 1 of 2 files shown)


openff/bespokefit/executor/services/qcgenerator/cache.py
```diff
@@ -50,6 +54,94 @@ def _canonicalize_task(task: _T) -> _T:
     return task
 
 
+def _sha256_hash(contents: str) -> str:
```
jthorton, line 57: As we are only using this to hash tasks does it make sense to make it a method of the base `QCGenerationTask`?

openff/bespokefit/executor/services/qcgenerator/cache.py
```diff
@@ -70,16 +162,13 @@ def cached_compute_task(
     # Canonicalize the task to improve the cache hit rate.
     task = _canonicalize_task(task)
```
jthorton, line 163: I think we need to do this first to maximise the cache hit rate, this will get skipped if the task is a `Torsion1DTask` as we return the method straight away.

openff/bespokefit/executor/services/qcgenerator/cache.py
```diff
@@ -50,6 +54,94 @@ def _canonicalize_task(task: _T) -> _T:
...
+
+def _compute_torsion_drive_task(
+    task: Torsion1DTask, redis_connection: redis.Redis
+) -> str:
+
+    task_hash = _sha256_hash(task.json())
+    task_id = _retrieve_cached_task_id(task_hash, redis_connection)
+
+    if task_id is not None:
+        return task_id
```
jthorton, line 89: I think its safe to assume if we are running this method then the full calculation is not in the cache. We can then use the logic in `cached_compute_task` to do this cache check?

openff/bespokefit/executor/services/qcgenerator/cache.py
jthorton: I think to work with my other comment above we could just reorder the function, I would do the canonicalize and cache look up first then the rest of the function as is after.
```suggestion
    task = _canonicalize_task(task)
    task_hash = _sha256_hash(task.json())
    task_id = _retrieve_cached_task_id(task_hash, redis_connection)
    
    if task_id is not None:
        return task_id
       
    if isinstance(task, Torsion1DTask):
        compute = worker.compute_torsion_drive
        return _compute_torsion_drive_task(task, redis_connection)
    elif isinstance(task, OptimizationTask):
        compute = worker.compute_optimization
    elif isinstance(task, HessianTask):
        compute = worker.compute_hessian
    else:
        raise NotImplementedError()
        
    task_id = compute.delay(task_json=task.json()).id
    _cache_task_id(task_id, task.type, task_hash, redis_connection)
    return task_id
```

SimonBoothroyd: > I think we need to do this first to maximise the cache hit rate, this will get skipped if the task is a Torsion1DTask as we return the method straight away.

Ah good catch - I missed this after shuffling the logic around!

openff/bespokefit/executor/services/qcgenerator/cache.py
SimonBoothroyd: This may be a good shout, although in the future we may need to hash other types such as internal intermediates depending on how the schemas are changed to for example allow pre-optimization. I'll leave it for now in this PR but feel free to make this change in #167!

PR #191, opened 2022-06-29
Issue templates

mattwthompson:
I yoinked these from ones I wrote for Interchange; you can roughly see what this would look like live from the pick-your-template page [there](https://github.com/openforcefield/openff-interchange/issues/new/choose).

Separating out feature requests, actual bug reports, and general user support fits well in my mental model of interacting with a user base, but these can easily be adapted to other structures. The only detail I'll strongly advocate for is making sure stuff like `conda list` is in the prompt for users that follow the bug report or user support templates.

I haven't actually used BespokeFit enough to know if files on disk are often used as relevant inputs or outputs, most stuff I've seen can be reproduced with a single command. We can add "please attach files" etc. to these templates if that's likely to be useful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/FEATURE_REQUEST.md
+++ b/.github/ISSUE_TEMPLATE/FEATURE_REQUEST.md
@@ -5,4 +5,5 @@ about: Suggest an improvement to BespokeFit
 ---
 
 **Description**
+
 Please describe the behavior you would like added to BespokeFit.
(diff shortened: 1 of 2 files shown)


j-wags (approved):
This is a clear improvement - I dumped a lot of work on @Yoshanuikabundi yesterday so I don't know if he'll get to this any time soon. Since this is a clear improvement I'd be happy for it to be merged here and refined in a subsequent PR.

Issue #189, opened 2022-06-19
Failure report - terrible fits for some small molecules with two opposite formal charges


simonlichtinger:
Dear bespoke fits devs,

### The setup

Installed as in previous issue reports, eg. #187 

Bespoke executor was run with

``` 
BEFLOW_OPTIMIZER_KEEP_FILES=True openff-bespoke executor launch  --n-fragmenter-workers 2   --n-optimizer-workers  12      --n-qc-compute-workers 6    --qc-compute-n-cores   2 --directory bespoke-tmp
``` 

And submissions were made using:
``` 
BEFLOW_OPTIMIZER_KEEP_FILES=True openff-bespoke executor submit  --file   "$lig"  --workflow-file  "workflow_$(basename $lig ).json" 
``` 

Where the workflow file was generated in python:
``` 
from openff.bespokefit.workflows import BespokeWorkflowFactory
from openff.toolkit.topology import Molecule
from openff.qcsubmit.common_structures import QCSpec, PCMSettings
from openff.bespokefit.executor import BespokeExecutor, BespokeWorkerConfig
from openff.bespokefit.executor import wait_until_complete
from openff.bespokefit.schema.optimizers import ForceBalanceSchema

molecules = ["ligands/alafosfalin.sdf", "ligands/aminolevulinicacid.sdf", "ligands/amoxicillin.sdf", "ligands/valacyclovir.sdf"]

for lig in molecules:
    factory = BespokeWorkflowFactory()
    ala_ala = Molecule.from_file(lig)
    qcspec = QCSpec(method='HF',basis='6-31G*',program='psi4')
    factory.default_qc_specs = [qcspec]
    factory.optimizer = ForceBalanceSchema()
    factory.optimizer.max_iterations = 50
    factory.optimizer.step_convergence_threshold = 0.005
    factory.optimizer.objective_convergence_threshold = 0.005
    factory.optimizer.gradient_convergence_threshold = 0.005
    workflow = factory.optimization_schema_from_molecule(ala_ala)

    factory.to_file(f"workflow_{lig.split('/')[1]}.json")
``` 

I had made the optimisation more sensitive and allowed more iterations since some of my molecules were throwing errors that the optimisation didn't converge.

### The failure case

Several ligands fail, the simplest one is aminolevulinic acid:
[aminolevulinicacid.sdf.zip](https://github.com/openforcefield/openff-bespokefit/files/8935628/aminolevulinicacid.sdf.zip)

[...]

PR #193, opened 2022-07-18
Deal with connectivity changes in QC generation

Yoshanuikabundi:
## Description
This PR addresses #189 by causing a clear error or warning and appropriately discarding data when the connectivity of the molecule being fitted changes. The goal is for connectivity changes to log a warning and fall back to initial force field parameters.

I'm opening this PR for discussion but it's a work in progress. At the moment it seems to work - SMILES `CCCC` completes successfully while SMILES `[NH3+]CC[O-]` and aminolevulinicacid.sdf from #189 raise the desired exception.

## Todos
  - [x] Identify when connectivity changes occur in QC generation
  - [x] Raise an exception on connectivity changes
  - [ ] ~~Identify when connectivity changes occur before QC generation (fragmentation?)~~
  - [ ] ~~Warn user that connectivity changes have occurred~~
  - [ ] ~~Skip targets where connectivity changes occur instead of raising exception~~
  - [x] Tests

## Questions
- [ ] ~~What's the best way to log a warning without failing the run?~~
- [x] A close inspection of #189 reveals that in the example provided, and in our minimized reproduction, the input coordinates to the torsion run have already undergone proton transfer. Where in code does this happen? I'm guessing when the fragmentation happens, but haven't found it yet
- [x] What's the best way to test this? I'm thinking a unit test of `OptimizationStage._inject_bespoke_qc_data()` plus an integration test of a number of molecules known both to and not to undergo a proton transfer, but this might be quite slow?
- [ ] ~~Are there currently any integration tests that will at least succeed if no connectivity change occurs?~~

## Status
- [x] Ready to go

### DIFF
--- a/devtools/conda-envs/test-env.yaml
+++ b/devtools/conda-envs/test-env.yaml
@@ -41,6 +41,7 @@ dependencies:
   - chemper
   - geometric
   - torsiondrive
+  - pymbar <4
 
     # Executor
   - uvicorn
--- a/devtools/conda-envs/test-env.yaml
+++ b/devtools/conda-envs/test-env.yaml
@@ -22,7 +22,7 @@ dependencies:
   - click-option-group
   - rdkit
   - openff-utilities
-  - openff-toolkit-base >=0.10.0,!=0.10.3
+  - openff-toolkit-base >=0.10.0,!=0.10.3,<0.11
   - openff-forcefields
   - openff-qcsubmit
   - openmm >=7.6.0
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union
 
 import httpx
-from openff.fragmenter.fragment import Fragment, FragmentationResult, Molecule
+from openff.fragmenter.fragment import Fragment, FragmentationResult
 from openff.toolkit.typing.engines.smirnoff import (
     AngleHandler,
     BondHandler,
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -15,7 +15,6 @@ from openff.toolkit.typing.engines.smirnoff import (
 )
 from pydantic import Field
 from qcelemental.models import AtomicResult, OptimizationResult
-from qcelemental.molutil import guess_connectivity
 from qcelemental.util import serialize
 from qcengine.procedures.torsiondrive import TorsionDriveResult
 from typing_extensions import Literal
@@ -533,34 +532,6 @@ class OptimizationStage(_Stage):
             )
 
             target.reference_data = local_qc_data
-
-            # Check that connectivity still matches the input molecule
-            for qc_record in local_qc_data.qc_records:
-                for name, qcschema in qc_record.final_molecules.items():
-                    fragment = Molecule.from_qcschema(qcschema)
-
-                    expected_connectivity = {
-                        tuple(sorted([bond.atom1_index + 1, bond.atom2_index + 1]))
-                        for bond in fragment.bonds
-                    }
-
-                    actual_connectivity = {
-                        tuple(sorted([a + 1, b + 1]))
-                        for a, b in guess_connectivity(
-                            qcschema.symbols, qcschema.geometry
-                        )
-                    }
-
-                    if expected_connectivity != actual_connectivity:
-                        raise RuntimeError(
-                            f"Target {qcschema.schema_name} record {name}: "
-                            + "Reference data does not match target.\n"
-                            + f"Expected mapped SMILES: {fragment.to_smiles(mapped=True)}\n"
-                            + "The following connections were expected but not found: "
-                            + f"{expected_connectivity - actual_connectivity}\n"
-                            + "The following connections were found but not expected: "
-                            + f"{actual_connectivity - expected_connectivity}\n"
-                        )
 
         targets_missing_qc_data = [
             target
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -532,6 +532,7 @@ class OptimizationStage(_Stage):
             )
 
             target.reference_data = local_qc_data
+            target.validate_reference_data()
 
         targets_missing_qc_data = [
             target
--- a/openff/bespokefit/executor/services/coordinator/stages.py
+++ b/openff/bespokefit/executor/services/coordinator/stages.py
@@ -15,6 +15,7 @@ from openff.toolkit.typing.engines.smirnoff import (
 )
 from pydantic import Field
 from qcelemental.models import AtomicResult, OptimizationResult
+from qcelemental.molutil import guess_connectivity
 from qcelemental.util import serialize
 from qcengine.procedures.torsiondrive import TorsionDriveResult
 from typing_extensions import Literal
@@ -533,18 +534,32 @@ class OptimizationStage(_Stage):
 
             target.reference_data = local_qc_data
 
-            from qcelemental.molutil import guess_connectivity
-
             # Check that connectivity still matches the input molecule
-            original_mol: Molecule = Molecule.from_smiles(input_schema.smiles)
             for qc_record in local_qc_data.qc_records:
-                for name, mol in qc_record.final_molecules.items():
-                    mol = Molecule.from_qcschema(mol)
-                    if not original_mol.is_isomorphic_with(mol):
+                for name, qcschema in qc_record.final_molecules.items():
+                    fragment = Molecule.from_qcschema(qcschema)
+
+                    expected_connectivity = {
+                        tuple(sorted([bond.atom1_index + 1, bond.atom2_index + 1]))
+                        for bond in fragment.bonds
+                    }
+
+                    actual_connectivity = {
+                        tuple(sorted([a + 1, b + 1]))
+                        for a, b in guess_connectivity(
+                            qcschema.symbols, qcschema.geometry
+                        )
+                    }
+
+                    if expected_connectivity != actual_connectivity:
                         raise RuntimeError(
-                            f"Target {input_schema.smiles} record {name}: Reference"
-                            + f" data connectivity does not match original molecule.\n"
-                            + f"{mol.to_smiles()} != {original_mol.to_smiles()}"
+                            f"Target {qcschema.schema_name} record {name}: "
+                            + f"Reference data does not match target.\n"
+                            + f"Expected mapped SMILES: {fragment.to_smiles(mapped=True)}\n"
+                            + f"The following connections were expected but not found: "
+                            + f"{expected_connectivity - actual_connectivity}\n"
+                            + f"The following connections were found but not expected: "
+                            + f"{actual_connectivity - expected_connectivity}\n"
                         )
 
         targets_missing_qc_data = [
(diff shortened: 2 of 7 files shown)


jthorton:
Thanks for taking this on @Yoshanuikabundi, just following along to make sure I understand whats happening and was curious about this point: 
> A close inspection of https://github.com/openforcefield/openff-bespokefit/issues/189 reveals that in the example provided, and in our minimized reproduction, the input coordinates to the torsion run have already undergone proton transfer. Where in code does this happen? I'm guessing when the fragmentation happens, but haven't found it yet

Where were you checking for this? I ran the following minimal code and found that the fragment molecule was the same as the input for the minimal example.
```python
from openff.toolkit.topology import Molecule
from openff.fragmenter.fragment import WBOFragmenter
mol = Molecule.from_smiles("[NH3+]CC[O-]")
w = WBOFragmenter()
result = w.fragment(mol)
# should get one fragment which is the same as the input 
result.fragment_molecules[(1, 2)] == mol
> True
```

Using fragmenter to image the result also confirms this 
image

openff/bespokefit/executor/services/coordinator/stages.py
```diff
@@ -530,6 +533,20 @@ async def _inject_bespoke_qc_data(
 
             target.reference_data = local_qc_data
 
+            from qcelemental.molutil import guess_connectivity
+
+            # Check that connectivity still matches the input molecule
+            original_mol: Molecule = Molecule.from_smiles(input_schema.smiles)
```
jthorton, line 539: For the case where the fragmentation stage produces different molecules we will need to use them as the reference as this is currently compared to the input molecule.  This causes an error to be raised even though the calculation runs fine. Confirmed to break with `C1=CC=C(C=C1)C2=CC=CC(=C2)C#N`.

Yoshanuikabundi:
Hey Josh, thanks for taking a look at this.

The reproducing code Jeff posted is this:

```
BEFLOW_OPTIMIZER_KEEP_FILES=True openff-bespoke executor run --smiles "[NH3+]CC[O-]" \
 --workflow           "default"    \
 --output             "xxx.json" \
 --output-force-field "xxx.offxml"  \
 --n-qc-compute-workers 2  \
 --qc-compute-n-cores   1  \
 --default-qc-spec xtb gfn2xtb none \
 --directory bespoke-tmp
```

For the run I just did, the `bespoke-tmp/35eed840-cede-46f2-9bdf-e30f173936a6/tmpi51l5oqg/stage_0/targets/torsion-0/conf.pdb` and `bespoke-tmp/35eed840-cede-46f2-9bdf-e30f173936a6/tmpi51l5oqg/stage_0/targets/torsion-0/input.sdf` files both have changed connectivity, and that changed connectivity is preserved during the torsion drive. In the [output files](https://github.com/openforcefield/openff-bespokefit/files/8935628/aminolevulinicacid.sdf.zip) posted by the user in #189, the same files for torsion-1 had changed connectivity.

I guess now's as good a time as any to ask - where does the optimization that generates those input files happen?

openff/bespokefit/executor/services/coordinator/stages.py
Yoshanuikabundi: Ohhhh yeah good point. OK.

openff/bespokefit/executor/services/coordinator/stages.py
```diff
@@ -530,6 +533,20 @@ async def _inject_bespoke_qc_data(
...
+            for qc_record in local_qc_data.qc_records:
+                for name, mol in qc_record.final_molecules.items():
+                    mol = Molecule.from_qcschema(mol)
```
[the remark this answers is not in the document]
Yoshanuikabundi: Interesting. I thought I got it to work as expected on a positive and negative test case (both with only one fragment). I remember being very surprised that it did what I expected 😅 I'll write a test and double check tomorrow.

I initially tried to use `ConnectivityFilter`, but I couldn't find a place where I could apply it as widely as here - I found one place that used a result collection but was concerned that it would have to be re-implemented every time someone added a qcgeneration type. I'll try `guess_connectivity` here again.

I raised an issue about writing a developer's guide because I was finding it really hard to figure out what part of the code I should be editing to do what I want.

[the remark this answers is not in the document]
jthorton: Yeah I think we might have to implement something similar here rather than use qcsubmit directly and putting it here should mean it works for all qc data types in future as they should all return a molecule and qcschema which we can compare. 

Yeah, great idea about the developer guide ill have to refresh myself 😅  and get back to your points!

Yoshanuikabundi:
Sorry @jthorton I've been distracted; I think I've got this working with `guess_connectivity`, and as a bonus the error message can be very descriptive:

```
Failed to inject the bespoke QC data into the optimization schema: Target qcschema_molecule record -120: Reference data does not match target.
Expected mapped SMILES: [H:1][C:2]([H:3])([C:4](=[O:5])[C:6]([H:7])([H:8])[N+:9]([H:10])([H:11])[H:12])[C:13]([H:14])([H:15])[C:16](=[O:17])[O-:18]
The following connections were expected but not found: {(9, 10)}
The following connections were found but not expected: {(10, 17)}
```

I was thinking this check could be appropriate as a validator on `TorsionProfileTargetSchema.reference_data`. That way it would run at the same time as this solution, but might be organised a bit better. A pro of this would be that it would avoid breaking target schemas like `VibrationTargetSchema` that don't have `final_molecule` attributes, and it would make tests very simple; a con would be that every new `TargetSchema` would have to re-implement it. Could get around this by making it part of an ABC. Thoughts?

I also am not sure how to test this without running multiple full bespoke fittings. I assume there's all sorts of test harnesses ready to use, but a primer on these would be great!

j-wags:
@jthorton Any tips on how to implement a test for connectivity rearrangement? I think @Yoshanuikabundi's architecture/base class question is nonessential, but the last thing before we open this for review is getting a test of the new behavior running.

Yoshanuikabundi:
OK I've written tests and moved everything to a validator on the various target schemas. I've also added support for result collections, in addition to local QC data, and for geometry optimizations in addition to torsion drives.

If we just want to stop a bespoke fitting when a connectivity change occurs in the reference data, I think we're ready for review! I've confirmed that the validation error stops the run but allows the executor to continue its other jobs, so no-one is going to lose a whole lot of good work because one molecule caused a connectivity change. If we want to try to salvage the good data when only one angle in the torsion drive is problematic, that might need to come in a future PR.

I haven't marked this PR as closing the raising issue (#189) because the requested fix was to add implicit solvent support, which this PR does not do.

devtools/conda-envs/test-env.yaml
```diff
@@ -41,6 +41,7 @@ dependencies:
   - chemper
   - geometric
   - torsiondrive
+  - pymbar <4
```
j-wags, line 44: Great catch. Hopefully this should be handled by the pins in the new ForceBalance build, but this is a good way to be sure.

j-wags (commented):
Nice! I'll leave the actual review to @jthorton but this looks quite good to me. 

This successfully raises the alarm for the minimum reproducing case in #189:

```
BEFLOW_OPTIMIZER_KEEP_FILES=True \
openff-bespoke executor run --smiles "[NH3+]CC[O-]" \
--workflow           "default"    \
 --output             "xxx.json" \
 --output-force-field "xxx.offxml"  \
 --n-qc-compute-workers 2  \
 --qc-compute-n-cores   1  \
--default-qc-spec xtb gfn2xtb none \
--directory bespoke-tmp
```

yields

```
[✓] qc-generation successful
[x] optimization failed
                                                                            
 "Failed to inject the bespoke QC data into the optimization schema: 1      
 validation error for TorsionProfileTargetSchema\nreference_data\n  Target  
 record -60: Reference data does not match target.\nExpected mapped SMILES: 
 [H:1][N+:2]([H:3])([H:4])[C:5]([H:6])([H:7])[C:8]([H:9])([H:10])[O-:11]\nTh
 e following connections were expected but not found: {(2, 3)}\nThe         
 following connections were found but not expected: {(3, 11)}\n             
 (type=value_error)"                                                        
                          
```

I do notice that, as opposed to a successful run, this doesn't make the bad conformations accessible in the `bespoke-tmp` directory, so if it's an easy fix it would be good to keep the temp files around in those cases. But if it would take more than 10 minutes then I think it's best to get this merged and cut a release, and that can be handled later.

j-wags:
@jthorton Do you think you'll have time to review this in the next week? If not I could be the primary reviewer.

openff/bespokefit/schema/targets.py
```diff
@@ -20,6 +23,140 @@
 from openff.bespokefit.utilities.pydantic import SchemaBase
 
 
+def _check_connectivity(
+    qcschema: QCEMolecule,
+    name: str,
+    fragment: Molecule,
+    expected_connectivity: Optional[Set[Tuple[int, int]]] = None,
```
jthorton, line 30: Can we not use the connectivity in the qcschema molecule as our `expected_connectivity`? As its made from the input openff molecule [here](https://github.com/openforcefield/openff-toolkit/blob/817563a06d8c583a66bea74df278b1eddbaa3124/openff/toolkit/topology/molecule.py#L4316) it should always be correct and the same and it removes the need for the `_off_mol_to_connectivity` function.

jthorton:
Hi thanks for the nudge and sorry for the delay. 

I like this being a function that is reusable as this will make it very easy for the QC worker to use the function in some sort of error handling way. If we take a normal geometry optimization job for example it would generate n conformers and then begin optimizing each one then return the first which completes and passes the connectivity check. 

I agree with Jeff that it would be good to return the geometries in someway so users can confirm a connectivity rearrangement has happened and bespokefit hasn't made a mistake, maybe it would be best to not have this as a validator so we can still store the result but have some validate method which is on the base class and is called after the QC data is injected and raises the informative error. This should mean users will be able to get the geometries in the json but the error will still stop the workflow.

Yoshanuikabundi:
> Can we not use the connectivity in the qcschema molecule as our expected_connectivity?

Good point, I think this was an ill advised premature optimization on my part. I was trying to avoid recomputing things that are the same across fragments. Having the fragment might also be necessary for the details in the error message, coz we need a mapped smiles that maps to the correct atom numbering. I'll make sure I can produce the same error message from `qcschema`, and if I can, remove the fragment and expected_connectivity arguments.

> maybe it would be best to not have this as a validator so we can still store the result but have some validate method which is on the base class and is called after the QC data is injected and raises the informative error. This should mean users will be able to get the geometries in the json but the error will still stop the workflow.

I think it might be easier just to include the geometries in the error message, and write them to disk there? I'm hesitant to put this in the base class because it's not appropriate for QC calculations that don't perform geometry optimizations (like `VibrationTargetSchema`), and the level of polymorphism the function has is already very complex. I'll write the geometries to disk and print the path in the error message.

jthorton:
> coz we need a mapped smiles that maps to the correct atom numbering. 

Good point this should be in the extras field of the qcschema molecule.

> I'm hesitant to put this in the base class because it's not appropriate for QC calculations that don't perform geometry optimizations (like VibrationTargetSchema)

I see we could have targets implement an abstract validate method as I could imagine some sort of quality control on most QC calculations. Note that all current targets Vibration (Vibration should do a Hessian at the optimised geometry, just noticed its not finished yet  😅 ), Optimised geometry and torsiondrive all require a geometry optimisation and could share a common base with the same validate method?

openff/bespokefit/schema/targets.py
```diff
@@ -20,6 +23,130 @@
...
+
+    ValueError
+        If the connectivity does not match.
+    """
+    # If expected connectivity is not provided, compute it from the fragment
+    if fragment is None:
+        fragment = Molecule.from_qcschema(qcschema)
+
+    # Get expected connectivity from bonds
+    expected_connectivity = {
```
jthorton, line 63: I think the expected connectivity should be in the qcschema molecule which would save calculating each time.

Yoshanuikabundi: I think that's true for `LocalQCData` but not for `TorsionDriveRecord` etc

jthorton (approved):
Thanks for taking this on @Yoshanuikabundi, LGTM!

PR #198, opened 2022-10-04
Support OpenFF Toolkit v0.11+

Yoshanuikabundi:
## Description
Update BespokeFit to work with new versions of the OpenFF Toolkit

## Todos
  - [x] Add pins for Toolkit v0.11+
  - [x] Transition to OpenFF Units
  - [x] Update for breaking changes in 0.11+
  - [x] Check `optimize.err` for errors in force balance handler
  - [x] Set minimum versions of toolkit, qcsubmit, and qcengine once they are released (see status)

## Status
- [x] Requires https://github.com/openforcefield/openff-qcsubmit/pull/204 in a release on conda forge
- [x] Requires https://github.com/openforcefield/openff-toolkit/issues/1420 to be fixed in a release on conda forge
- [x] Requires https://github.com/leeping/forcebalance/pull/259 in a release on conda forge (passes tests without this)
- [x] New ForceBalance release
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -81,7 +81,7 @@ jobs:
       - name: Run Tests
         shell: bash -l {0}
         run: |
-          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests/ --cov-report=xml
+          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests/executor/services/coordinator/test_app.py::test_get_optimization --cov-report=xml
 
       - name: Codecov
         uses: codecov/codecov-action@v3.1.1
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -68,7 +68,7 @@ jobs:
       - name: Run Tests
         shell: bash -l {0}
         run: |
-          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests/executor/services/coordinator/test_app.py::test_get_optimization --cov-report=xml
+          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests --cov-report=xml
 
       - name: Codecov
         uses: codecov/codecov-action@v3.1.1
--- a/devtools/conda-envs/no_openeye.yaml
+++ b/devtools/conda-envs/no_openeye.yaml
@@ -22,7 +22,7 @@ dependencies:
   - rdkit >=22
   - ambertools >=22
   - openff-utilities
-  - openff-toolkit-base =0.11
+  - openff-toolkit-base >=0.11.3
   - openff-forcefields >=2.0.0
   - openff-qcsubmit >=0.3.0
   - openmm >=7.6.0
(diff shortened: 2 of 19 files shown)


j-wags (approved):
The molecule in `test_submit::test_submit_multi_molecule` is failing because RDKitToolkitWrapper thinks Cu2+ is a radical. So the toolkit should be changed to accept Cu2+ so that tests here pass. 

This also touches on the issue that QCSubmit sometimes puts multiple non-covalently-bound components in a single OFFMol, but while that's still something we need to get rid of some day, that's not the root cause here and so it shouldn't be a blocker to this PR. 

So, accepted once the tests pass (which will probably involve merging a Toolkit PR)

jthorton:
@Yoshanuikabundi I think this test should work with any multi-molecule system maybe something like `Cc1ccc(cc1Nc2nccc(n2)c3cccnc3)NC(=O)c4ccc(cc4)CN5CCN(CC5)C.CS(=O)(=O)O` imatinib mesylate would do the job to save you waiting on a toolkit PR?

Yoshanuikabundi:
@jthorton Thanks but the Toolkit PR is already in :)

Yoshanuikabundi:
Looks like we also need https://github.com/leeping/forcebalance/pull/259, which is currently in limbo. Tests for force balance pass, but trying to actually run an optimization fails. The error message is extremely cryptic and the true cause is difficult to lock down owing to a lack of `optimize.err` handling in `ForceBalanceOptimizer._read_output()` and the aggressive clean up of temporary files throughout the codebase. BespokeFit believes FB to have completed successfully when it actually errored out at an early stage, leading to a failure to find result outputs. I'll add in some code to handle this once I've got something ready for the BespokeFit workshop.

mattwthompson:
Tests are starting to come in with some green checks - if anybody wishes to once it over again I'll give a window of a few days before merging (and another window for checks prior to making/pushing for a release).

.github/workflows/CI.yaml
```diff
@@ -68,7 +68,7 @@ jobs:
       - name: Run Tests
         shell: bash -l {0}
         run: |
-          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests/ --cov-report=xml
+          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests/executor/services/coordinator/test_app.py::test_get_optimization --cov-report=xml
```
Yoshanuikabundi, line 71:
```suggestion
          pytest -v --cov=openff --cov-config=setup.cfg openff/bespokefit/tests --cov-report=xml
```

I think this was done to speed up tests? I'm going to remove it to make sure we still get the green check but documenting it in case its important.

devtools/conda-envs/no_openeye.yaml
```diff
@@ -63,8 +63,5 @@ dependencies:
   - codecov
   - requests-mock
 
-  # Temporary branches
-  - pip:
-    - git+https://github.com/openforcefield/openff-toolkit.git@main
-    - git+https://github.com/yoshanuikabundi/openff-qcsubmit.git@toolkit_011_support
-    - git+https://github.com/Yoshanuikabundi/QCEngine.git@off_toolkit_011_support
+  - importlib-metadata >=4
+  - importlib_metadata >=4
```
Yoshanuikabundi, line 67:
```suggestion
  - importlib-metadata >=4
```

Duplicate

mattwthompson: Sure, not a typo but no longer strictly necessary to include here

https://github.com/openforcefield/openff-bespokefit/pull/202#issuecomment-1314112832
https://github.com/conda-forge/openff-toolkit-feedstock/pull/37

PR #199, opened 2022-10-11
Add option to always keep temporary files

Yoshanuikabundi:
## Description
Trouble shooting bespoke fit is very frustrating, in part because temporary files are aggressively cleaned up. This PR attempts to add a setting, `BEFLOW_KEEP_TMP_FILES`, that ensures that temporary files are kept (though perhaps not in an obvious location)

## Todos
  - [x] Add setting
  - [x] Add new `temporary_cd()` implementation using setting
  - [x] Replace uses of `temporary_cd()`
  - [x] See if it works
  - [ ] Investigate other changes needed to keep files
  - [x] Try to keep temporary file creation inside the  requested working directory
  - [x] Rebase after #198 is merged

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/_settings.py
+++ b/openff/bespokefit/executor/services/_settings.py
@@ -62,6 +62,12 @@ class Settings(BaseSettings):
     BEFLOW_OPTIMIZER_WORKER_N_CORES: Union[int, Literal["auto"]] = "auto"
     BEFLOW_OPTIMIZER_WORKER_MAX_MEM: Union[float, Literal["auto"]] = "auto"
     BEFLOW_OPTIMIZER_KEEP_FILES: bool = False
+    """
+    .. deprecated:: 0.2.1
+        use BEFLOW_KEEP_TMP_FILES instead
+
+    Keep the optimizer's temporary files.
+    """
 
     BEFLOW_KEEP_TMP_FILES: bool = False
     """Keep all temporary files."""
--- a/openff/bespokefit/executor/services/_settings.py
+++ b/openff/bespokefit/executor/services/_settings.py
@@ -63,6 +63,9 @@ class Settings(BaseSettings):
     BEFLOW_OPTIMIZER_WORKER_MAX_MEM: Union[float, Literal["auto"]] = "auto"
     BEFLOW_OPTIMIZER_KEEP_FILES: bool = False
 
+    BEFLOW_KEEP_TMP_FILES = False
+    """Keep all temporary files."""
+
     @property
     def fragmenter_settings(self) -> WorkerSettings:
         return WorkerSettings(
--- a/openff/bespokefit/executor/services/_settings.py
+++ b/openff/bespokefit/executor/services/_settings.py
@@ -63,7 +63,7 @@ class Settings(BaseSettings):
     BEFLOW_OPTIMIZER_WORKER_MAX_MEM: Union[float, Literal["auto"]] = "auto"
     BEFLOW_OPTIMIZER_KEEP_FILES: bool = False
 
-    BEFLOW_KEEP_TMP_FILES = False
+    BEFLOW_KEEP_TMP_FILES: bool = False
     """Keep all temporary files."""
 
     @property
--- a/openff/bespokefit/utilities/tempcd.py
+++ b/openff/bespokefit/utilities/tempcd.py
@@ -1,0 +1,53 @@
+""""""
+import os
+import shutil
+import tempfile
+from contextlib import contextmanager
+from typing import Optional
+
+from openff.bespokefit.executor.services import current_settings
+
+
+@contextmanager
+def temporary_cd(path: Optional[str] = None, cleanup: Optional[bool] = None):
+    """Temporarily move the current working directory to the path specified.
+
+    Parameters
+    ----------
+    path
+        The path to CD into. If ``None`` or not specified, a temporary directory
+        will be created.
+    cleanup
+        Whether the directory and its contents should be kept after the context
+        manager exits. If ``True``, the temporary directory and its contents
+        will be destroyed when the context manager exits; if ``False``, it will
+        be kept. If ``None`` or not specified, the directory is destroyed if
+        ``Path`` is not specified and the BEFLOW_KEEP_TMP_FILES setting is
+        ``False``.
+    Returns
+    -------
+    """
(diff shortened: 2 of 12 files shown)


openff/bespokefit/executor/services/_settings.py
```diff
@@ -65,6 +65,9 @@ class Settings(BaseSettings):
     BEFLOW_OPTIMIZER_WORKER_MAX_MEM: Union[float, Literal["auto"]] = "auto"
     BEFLOW_OPTIMIZER_KEEP_FILES: bool = False
```
jthorton, line 66: @Yoshanuikabundi a similar option is already exposed here which causes the forcebalance files to be kept rather than cleaned up, happy to change to a more general name that includes all temp files but we probably don't need both?

jthorton:
see [here](https://github.com/openforcefield/openff-bespokefit/blob/5663b522fc766faa9b05a5765d0229f7ba2499f7/openff/bespokefit/executor/services/optimizer/worker.py#L89) for how this currently works I think this more general solution looks better and more useful so feel free to remove the old logic.

I wonder if it would be worth extending the results schema to include a `file_path` to the location of files so users can work out which directory corresponds to which optimisation?

openff/bespokefit/executor/services/_settings.py
Yoshanuikabundi: I agree but this would be a breaking change, and we use that environment variable a lot in demonstrations and troubleshooting. Maybe we should deprecate the old name and temporarily make the old name an alias of the new behaviour until we remove it?

Yoshanuikabundi:
Yeah I wrote the current implementation because I was frustrated with losing my files if the optimisation errored out before it gets to the copy step. I agree that my implementation is delightfully general and useful but unfortunately it has the small downside of not working 😅 The temporary directories themselves are preserved but not their contents. I'll keep working on it when I get a chance but I need to focus on the workshop at the moment.

mattwthompson:
Is this closer to 0.1.3 or 0.2.0? This might be very useful when working on the step that interacts with ForceBalance if it can bypass the need to generate data each run.

Yoshanuikabundi:
Yeah this will be super useful for debugging but it's a ways off. It ran into the usual problem - unit tests weren't capturing the bugs it introduced. So it's probably a post-0.2.0 feature.

The goal was not to do any caching, just to keep the files around - BespokeFit already has a caching feature for the QC step, it's probably simpler to extend that to optimisation.

Yoshanuikabundi:
I think that's the approach this PR takes? I (tried to) refactor the `temporary_cd` function to create a temporary directory, and then only clean it up if the appropriate setting is on. If the setting is off, the directory is created but not cleaned up. Naming things after the task id is a great idea but I've gotta get this working first :upside_down_face:

jthorton:
> I'd like to name folders after their IDs

Thats great I think this will help a lot, I think that we will have to change each workers call to the delay function to include a task id like the example [here](https://github.com/celery/celery/issues/1813#issuecomment-33142648), I think we should also have this in the task schema and results schema of each job so users can track the task as its running. Maybe these changes would be better handled in another PR as its changing the scope of this PR. Happy to take that on!

Yoshanuikabundi:
That makes sense to me! If you could take a crack at that, that'd be great - I think we should merge this, #239, and #241, release 0.2.1 next week, and then we can remove the old `BEFLOW_OPTIMIZER_KEEP_FILES` setting and name temporary folders after the task ID in the next release. Something something breaking changes something something lets just get a release that supports toolkit 0.11+ out.

If you could also formally review this PR that'd be great!

openff/bespokefit/executor/services/_settings.py
```diff
@@ -62,6 +62,15 @@ class Settings(BaseSettings):
...
+    """
+    .. deprecated:: 0.2.1
+        use BEFLOW_KEEP_TMP_FILES instead
+
+    Keep the optimizer's temporary files.
+    """
+
+    BEFLOW_KEEP_TMP_FILES: bool = False
```
jthorton, line 72: This might be a good time to introduce a name change or do we want to keep it consistent till we break all of these?

openff/bespokefit/utilities/tempcd.py
```diff
@@ -0,0 +1,55 @@
...
+    path
+        The path to CD into. If ``None`` or not specified, a temporary directory
+        will be created. If specified but the path does not exist, a temporary
+        directory with that name will be created.
+    """
+    # Normalize path to a pathlib Path
+    path: Optional[Path] = None if path is None else Path(path)
+
+    # Decide whether to clean up based on bespokefit settings
+    cleanup = not current_settings().BEFLOW_KEEP_TMP_FILES
```
jthorton, line 32: This should depend on both the old and new variables.

Yoshanuikabundi: I wanted to keep the new behavior and the old behavior separate for the same breaking things reasons as above... in 0.3.0 or whatever we'll deprecate/remove the old setting altogether, but for now I don't wanna break any scripts people have that assume the files are going to a particular place.

Yoshanuikabundi: Welp ok yep so these changes break the old setting anyway - since the source and destination for the files is the same, the optimizer worker cleans up the copied files [here](https://github.com/openforcefield/openff-bespokefit/blob/6b57aa6464879a0cebc58b87f227b613e08d4fdd/openff/bespokefit/executor/executor.py#L347-L348). So forget everything I said your way is perfect :)

openff/bespokefit/executor/services/_settings.py
Yoshanuikabundi: Oh I see, its a custom thing. Uhm yes I would like machinery like this to handle deprecations, no I don't want to put it in this PR :P

Yoshanuikabundi:
I shoulda just tried out the task ID thing before commenting everywhere 😅 I love how using the task ID consolidates all the logic - this'll make removing the old setting easy in the future. We just need to remember to tempcd to the task id whenever we write temporary files to disk.

I'll merge this on Monday unless you find something else to fix!

jthorton (approved):
Thanks for this @Yoshanuikabundi, looks fantastic!

PR #202, opened 2022-11-14
Use micromamba in CI

mattwthompson:

PR #203, opened 2022-11-15
Re-run CI

mattwthompson:
CI is failing on feature branches for reasons that are not clear to me. I'm re-running CI here to see if the failures are genuine and on the `main` branch.

### DIFF
--- a/devtools/conda-envs/no_openeye.yaml
+++ b/devtools/conda-envs/no_openeye.yaml
@@ -43,7 +43,7 @@ dependencies:
 
     # Executor
   - uvicorn
-  - fastapi
+  - starlette =0.20 # https://github.com/openforcefield/openff-bespokefit/pull/203#issuecomment-1315936000
   - celery
   - httpx
   - redis-server
--- a/devtools/conda-envs/no_openeye.yaml
+++ b/devtools/conda-envs/no_openeye.yaml
@@ -37,7 +37,7 @@ dependencies:
   - qcelemental >=0.24.0
   - qcengine >=0.23.0
   - chemper
-  - geometric
+  - geometric <1
   - torsiondrive
   - pymbar <4
 
--- a/devtools/conda-envs/no_openeye.yaml
+++ b/devtools/conda-envs/no_openeye.yaml
@@ -57,3 +57,6 @@ dependencies:
   - pytest-celery
   - codecov
   - requests-mock
+
+  - importlib-metadata >=4
+  - importlib_metadata >=4
--- a/devtools/conda-envs/no_openeye.yaml
+++ b/devtools/conda-envs/no_openeye.yaml
@@ -37,7 +37,7 @@ dependencies:
   - qcelemental >=0.24.0
   - qcengine >=0.23.0
   - chemper
-  - geometric <1
+  - geometric
   - torsiondrive
   - pymbar <4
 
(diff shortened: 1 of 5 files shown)


j-wags:
Oh, wow, thanks for catching this.

It looks like 4 errors have started to happen:
* The importlib thing which you caught here (and are fixing other places too)
* A bunch [errors around the pint 0.20 release](https://github.com/openforcefield/openff-bespokefit/actions/runs/3365705820/jobs/5581434307#step:8:3004) which got resolved 15 days ago by the openff-units release
* A [test](https://github.com/openforcefield/openff-bespokefit/actions/runs/3432848295/jobs/5722535316#step:8:471) which has been flaky for a while `openff/bespokefit/tests/executor/services/qcgenerator/test_worker.py::test_compute_torsion_drive` where an optimization doesn't converge and raises `geometric.errors.GeomOptNotConvergedError: Optimizer.optimizeGeometry() failed to converge.`
* Whatever this [new thing](https://github.com/openforcefield/openff-bespokefit/actions/runs/3472346095/jobs/5803051115#step:8:421) is, in `test_get_optimization` which manifests as `HTTPStatusError`. I agree with you that it only seems to manifests on PRs from branches in the last few days. 

So I'll keep looking into this last one, and I may work on that flaky convergence one as well.

j-wags:
Fastapi is a possible culprit here, though they have a really fast release pace so maybe it's nothing. https://anaconda.org/conda-forge/fastapi/files

mattwthompson:
Here's the most recent nightly CI run the passed: https://github.com/openforcefield/openff-bespokefit/actions/runs/3231103232/jobs/5290258911

j-wags:
Looks like that was it. Possible even rootier root issue is the [fastapi conda package](https://anaconda.org/conda-forge/fastapi/files) changing from requiring starlette 0.20 to 0.21. Rabbit hole begins here: https://github.com/tiangolo/fastapi/issues/5646

mattwthompson:
^ above failures vs. passing tests is because I forgot to update both environments, not necessarily because of some mystical relationship between OpenEye and Starlette

mattwthompson:
If this pin works, my current plan is:

* Slim down this PR to relevant changes, merge
* Update #202, merge
* Merge changes into #198

I can also push the pin into the conda builds if you think that's worthwhile

devtools/conda-envs/no_openeye.yaml
```diff
@@ -37,13 +37,13 @@ dependencies:
...
   - chemper
-  - geometric
+  - geometric <1
   - torsiondrive
   - pymbar <4
 
     # Executor
   - uvicorn
-  - fastapi
+  - starlette =0.20 # https://github.com/openforcefield/openff-bespokefit/pull/203#issuecomment-1315936000
```
j-wags, line 46:
```suggestion
  - fastapi
  - starlette =0.20 # https://github.com/openforcefield/openff-bespokefit/pull/203#issuecomment-1315936000
```

j-wags:
That plan sounds good to me, and I agree we should pin the conda builds as well! 

I'd like to gain a better understanding of where the rift in our fitting stack is going to open up - There's some range of packages that work with OFFTK 0.10, and another that works with OFFTK 0.11. 

Also I'm a little bothered by the sporadic GeomeTRIC convergence errors, but I'll open a separate issue for that.

mattwthompson:
Merging these fixes as part of #202 since these environments are pretty slow to solve (one took 50 minutes)

PR #204, opened 2022-12-12
Fix linting workflow

mattwthompson:
This might not be necessary in a few hours or days


mattwthompson:
This has been fixed upstream; `isort` 5.11.0 is broken for this use but 5.11.1 works

PR #208, opened 2022-12-15
Test on Python 3.9-3.10

mattwthompson:
Just to see what happens ...


mattwthompson:
The macOS builds each fail (on 3.10 because OpenMM is not available and on other versions because of gateway timeouts). The Linux builds each seem to pass.

PR #212, opened 2023-01-06
Update Psi4 installation instructions

### LINKED ISSUE #210: Installation instructions unable to install new versions of OpenFF Toolkit and Psi4
[This](https://docs.openforcefield.org/projects/bespokefit/en/stable/getting-started/installation.html#psi4) pulls down version 0.10.6:

```console
$ CONDA_SUBDIR=osx-64 mamba create -n bespokefit-env -c psi4 -c conda-forge -c defaults -c openeye python=3.9 openff-bespokefit psi4 openeye-toolkits "h5py<3.2"
```

which is no longer maintained short of critical bugfixes. Using instead `h5py >3.2` in the above command works in the sense that it finds a solution quickly, but does not work in the sense that it brings down a version of the OpenFF stack based on 0.10.6. I believe https://github.com/conda-forge/openff-bespokefit-feedstock/commit/4cf88a60579c630c6d587290515ae215a179e098 is related and should be reverted (https://github.com/conda-forge/openff-bespokefit-feedstock/pull/9).

The [standard quick-start installation instructions](https://docs.openforcefield.org/projects/bespokefit/en/stable/getting-started/quick-start.html) pull down version 0.11.4, so I don't think the problem is necessarily widespread.

mattwthompson:
Resolves #210

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -58,18 +58,18 @@ create a working conda environment containing both.
 :::
 
 :::{note}
-On Linux, installing Psi4 into an existing environment sometimes fails because of subtle differences in
+Installing Psi4 into an existing environment sometimes fails because of subtle differences in
 compiled dependencies found in multiple channels. An alternative is to install everything when
 initially creating the environment using, with AmberTools:
 
 ```shell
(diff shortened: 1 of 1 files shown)


j-wags (approved):
LGTM. IIRC this was just an emergency docs patch to make sure the bespokefit followup workshop would go smoothly.

Issue #218, opened 2023-01-19
Integration tests!


Yoshanuikabundi:
Let's add an integration test suite to BespokeFit.

BespokeFit has good unit-test coverage, but much of its behavior is non-local. The 0.2.0 release was broken because of a bug in ForceBalance which was not detected by unit tests. The goal of this test suite should be to provide a place to test for non-local interactions in the codebase, to catch bugs before they are released, and to ensure that deployed versions of BespokeFit are still working.

The integration test suite should test at least the following behavior:
- [ ] BespokeFit can complete a torsion optimization of a simple, drug-like molecule (aspirin? one of the molecules we report in the paper?) using XTB (and otherwise the default settings) without raising an error and with reasonable output torsions.
- [ ] BespokeFit can complete a torsion optimization of ethane, hydrogen peroxide, or a similarly simple molecule with the default settings (using Psi4, and except for necessary changes to `torsion_smirks`) without raising an error.
- [ ] A failure in ForceBalance leading to a traceback in `optimise.err` is correctly reported (#217, #216, https://github.com/openforcefield/openff-bespokefit/pull/198#issuecomment-1274434924)
- [ ] (more to add later)

This integration suite should not be run in PRs (it'll be too slow), but it should be executed:
- [ ] Before releases
- [ ] Immediately after releases (with the deployed environment)
- [ ] After a PR is merged (or a commit is made to `main`)
- [ ] If it hasn't been run in the last week (to catch any changes released upstream)

This integration suite should pull in dependencies from Conda as much as possible to match the recipe for the next release.

This integration suite should NOT be derived from the documentation - while code in the documentation should be tested, that is not the purpose of this suite. The documentation should be free to change without considering the impact on test coverage, and it should be possible to add to the test suite without having to invent new documentation.

jthorton:
Just adding some notes on running the current integration test locally. 
Created a `bespokefit-0.1.3`  environment on mac using the command 
`mamba create --name bespokefit-1-3 openff-bespokefit=0.1.3 openff-toolkit  xtb-python`

<details><summary>Conda list output</summary>
<p>
# packages in environment at /Users/joshua/mambaforge/envs/bespokefit-1-3:
#
# Name                    Version                   Build  Channel
amberlite                 22.0                     pypi_0    pypi
ambertools                22.2            py310h6a0e7a1_0    conda-forge
amberutils                21.0                     pypi_0    pypi
[...]

PR #222, opened 2023-01-26
Add a barebones integration test to CI

Yoshanuikabundi:
## Description
This is a barebones integration test for CI. It just runs the following fit:

```shell
openff-bespoke executor run --smiles                 'CC'               \
                            --workflow               'default'          \
                            --default-qc-spec        xtb gfn2xtb none   \
                            --target-torsion         '[C:1]-[C:2]'
```

This requires adding `xtb-python` to the test environment.

Integration tests are added to the testing matrix so that existing unit tests don't take longer; instead, there are 4 additional checks. Once I see what the timing is like on this PR, I'll probably adjust this.

This PR partially mitigates #218 but does not close it.

### DIFF
--- a/devtools/conda-envs/test-env.yaml
+++ b/devtools/conda-envs/test-env.yaml
@@ -35,7 +35,7 @@ dependencies:
     # Optional
   - forcebalance
   - openff-fragmenter-base
-
+  - xtb-python
   - openeye-toolkits
 
     ### Bespoke dependencies
--- a/devtools/conda-envs/test-env.yaml
+++ b/devtools/conda-envs/test-env.yaml
@@ -24,6 +24,7 @@ dependencies:
   - openff-utilities
   - openff-toolkit-base >=0.11.3
   - openff-forcefields
+  - openff-interchange
   - openff-units
   - openff-qcsubmit
   - openmm >=7.6.0
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -22,7 +22,7 @@ jobs:
         os: [ubuntu-latest, macOS-latest]
         python-version: [3.8]
         openeye: [true, false]
-        integration: [true, false]
+        integration: [true]
 
     env:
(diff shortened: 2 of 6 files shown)


Yoshanuikabundi:
Only 4 checks seem to be running at a time (the other 4 are queued), so as long as the integration fit takes less than a few minutes (after environment creation) it's probably worth just adding it to the same job as everything else.

mattwthompson:
image

Hey, that actually is pretty quick

Yoshanuikabundi:
Yeah! I think we can afford to have that run on all our CI.

Yoshanuikabundi:
OK the integration test is failing on forcebalance. Which is good, because forcebalance is broken!

devtools/conda-envs/test-env.yaml
```diff
@@ -35,7 +36,7 @@ dependencies:
     # Optional
   - forcebalance
   - openff-fragmenter-base
-
+  - xtb-python
```
mattwthompson, line 39: This is the only change I'm iffy on (non-blocking). I think it's pretty light but it's only used in ~half of the tests, being dead weight in the others.

mattwthompson (approved):
I think this should be merged in its current state, i.e. unit tests passing and the new integration test failing (matching the actual behavior you get trying to use 0.2.0/1.9.4)

devtools/conda-envs/test-env.yaml
Yoshanuikabundi: Yeah, without it we'd need to either add a new environment for integration tests or write some sort of test QC generation harness, neither of which I want to do.

MacOS/RDKit failed a couple of unit tests. I'm re-running it but I'll probably merge even with the unit test failures, coz I'm not sure how this PR could've caused them. Completely agree with you about having the integration test fail.

mattwthompson: I also see reliably sporadic unit test failures in this codebase (and don't know how to fix them)

PR #225, opened 2023-02-01
Reformat with Black 23.1

mattwthompson:
This PR reformats the codebase with Black 23.1, which notably removes blank lines after function definitions and some conditionals, a pattern Simon used everywhere and one I like.

I'm opening this now because every PR will fail without this change since the linting workflow enforces a requirement that the most recent version of every tool must pass. Using `pre-commit` would allow for more explicit dependency management and non-zero control over the pace of these upgrades.


jthorton:
Thanks for opening this! I am not a fan of this change and find the old code more readable personly, I am not that familiar with `pre-commit` yet but it seems great and I would be all for adding it bespokefit!

mattwthompson:
I'm also not a fan of these changes - but it's one of the things we give up when using opinionated tools.

I'll spin up `pre-commit` in another PR; I like it but I don't know if other maintainers do or not

Yoshanuikabundi (approved):
I'm into it! I think there's value in using the same formatting as everyone else, even though there are some things about Black that drive me nuts. 

Before you merge this, could you create a [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) file with the hash of the commit in which you ran Black? That way this change gets ignored when computing blames in GitHub.

I also LOVE the pre-commit setup you have on Interchange where it runs automatically on every commit. If you could set that up here I'd be super grateful! I've been meaning to get around to it. Adding lints for notebook formatting and docstrings is somewhere on my list too.

mattwthompson:
Nice, I didn't know about that feature.

I'm not seeing 85d5238 in one of the blames I pulled up so I'll merge this when tests pass/pass-ish.

https://github.com/openforcefield/openff-bespokefit/blame/71104d274591d5db48d6c134c58e1d9bd2fbd56a/openff/bespokefit/__init__.py

mattwthompson:
The failing tests match what is failing on the current `main` branch so I don't think there's any reason to believe these changes are the cause.

Issue #234, opened 2023-03-02
Failing tests


mattwthompson:
Just reporting these, I won't have the time to fix them. They seem important.

```
FAILED openff/bespokefit/tests/cli/test_cache.py::test_update_from_qcsubmit - redis.exceptions.DataError: Invalid input of type: 'NoneType'. Convert to a bytes, string, int or float first.
FAILED openff/bespokefit/tests/executor/services/qcgenerator/test_cache.py::test_canonicalize_torsion_task - AssertionError: assert '[H:2][C:1]([H])([H])O[H]' == '[H][C:1]([H])([H])[O:2][H]'
  - [H][C:1]([H])([H])[O:2][H]
  ?                   - ---
  + [H:2][C:1]([H])([H])O[H]
  ?   ++
FAILED openff/bespokefit/tests/utilities/test_molecule.py::test_canonical_order_atoms[with_rd_backend] - assert (3, 2) in [(2, 5), (5, 2)]
FAILED openff/bespokefit/tests/utilities/test_molecule.py::test_canonical_order_atoms[with_oe_backend] - assert (5, 4) in [(2, 5), (5, 2)]
```

https://github.com/openforcefield/openff-bespokefit/actions/runs/4309177826/jobs/7516259466

PR #235, opened 2023-03-03
Fix #237

### LINKED ISSUE #237: The executor fails with multiple ligands
Dear members of the bespokeFit community,

I need your help/advice. I have a problem processing multiple ligands (ligand sets in parallel) with BespokeFit. Jobs often fail, usually at the Optimization stage (see the error message below). The failed task can usually be finished if the task is restarted, or if the ligand is processed as a single BespokeFit job. I tested  BespokeFit with the fep-benchmark set and with proprietary ligands, mostly with the semiempirical xtb code, and few with psi4.

I follow the @jthorton 's advice given here previously (https://github.com/openforcefield/openff-bespokefit/issues/215).

I have a feeling that the issue has something to do with memory management. I gave the scripts I use to a colleague of mine and his jobs failed multiple times on a Linux machine he uses as a personal computer. 

I would appreciate any advice you can share.

I run BespokeFit with 1 or 2 cores (with psi4 jobs I use 4):

```
with BespokeExecutor(
    n_fragmenter_workers = 12,
    n_optimizer_workers = 40,
    n_qc_compute_workers = 40,
    qc_compute_worker_config=BespokeWorkerConfig(n_cores=2)
) as executor:
    # Submit our workflow to the executor
    ids = []
    for optimization_schema in workflow_schema:
        task_id = executor.submit(optimization_schema)
        ids.append(task_id)

```

When the jobs are done I extract individual FFs with the next code:

```
import sys
from openff.bespokefit.executor import BespokeExecutor
import pandas as pd
from simtk import unit


id = sys.argv[1]
output = BespokeExecutor.retrieve(optimization_id=id)

print(output.status)
print(output.error)
print(id)
print("====================================================")


if output.error is None:
   force_field = output.bespoke_force_field
   force_field.to_file("force_field_from_QMM_id_{0}.offxml".format(id))

   print(getattr(output.results.input_schema, 'target_torsion_smirks'))
  
   parameter_data = []
   for i, (parameter, initial_values) in enumerate(output.results.input_schema.initial_parameter_values.items()):
      for final_parameter, final_values in output.results.refit_parameter_values.items():
         if parameter.smirks == final_parameter.smirks:
            print(parameter.smirks) 
            for term in range(1, 5):
              k_before = initial_values[f"k{term}"].value_in_unit(unit.kilocalorie_per_mole)
              k_after = final_values[f"k{term}"].value_in_unit(unit.kilocalorie_per_mole)
              parameter_data.append([f"smirks_{i}_k{term}", k_before, k_after, k_after - k_before])

   df = pd.DataFrame(parameter_data, columns=["parameter", "before", "after", "change"])
   print(df)
   df.to_csv("optimized_parameters_id_{}.csv".format(id),index=False)
```

and merge them with

```
rm cffs.sh
echo "openff-bespoke combine --output \"combined_bespokefit_forcefield_for_cdk8_cmet_eg5_pfkfb3.offxml\"  \\"|tee cffs.sh

for filename in force*.offxml; do
	echo $filename
	echo "--ff \"${filename}\"               \\" &>> cffs.sh 
done

chmod +x cffs.sh
./cffs.sh
```

<details><summary>Error message:</summary><pre>
{"type": "RuntimeError", "message": "ConvergenceFailure: The optimization failed to converge.\nNone", "traceback": "Traceback (most 
 recent call last):\n  File \"/home/ognjen/anaconda3/envs/bespokefit/lib/python3.8/site-packages/celery/app/trace.py\", line 451, in 
 trace_task\n    R = retval = fun(*args, **kwargs)\n  File                                                                           
 \"/home/ognjen/anaconda3/envs/bespokefit/lib/python3.8/site-packages/celery/app/trace.py\", line 734, in __protected_call__\n       
 return self.run(*args, **kwargs)\n  File                                                                                            
 \"/home/ognjen/anaconda3/envs/bespokefit/lib/python3.8/site-packages/openff/bespokefit/executor/services/optimizer/worker.py\", line
 74, in optimize\n    raise (\nRuntimeError: ConvergenceFailure: The optimization failed to converge.\nNone\n"}   
</pre></details>

<details><summary>Hardware:</summary><pre>
Architecture:                    x86_64
CPU op-mode(s):                  32-bit, 64-bit
Byte Order:                      Little Endian
Address sizes:                   46 bits physical, 48 bits virtual
CPU(s):                          94
On-line CPU(s) list:             0-93
Thread(s) per core:              1
Core(s) per socket:              94
Socket(s):                       1
NUMA node(s):                    1
Vendor ID:                       GenuineIntel
CPU family:                      6
Model:                           85
Model name:                      Intel(R) Xeon(R) Platinum 8168 CPU @ 2.70GHz
Stepping:                        4
CPU MHz:                         2693.670
BogoMIPS:                        5387.34
Virtualization:                  VT-x
L1d cache:                       3 MiB
L1i cache:                       3 MiB
L2 cache:                        376 MiB
L3 cache:                        16 MiB
NUMA node0 CPU(s):               0-93
....
</pre></details>

<details><summary>OS:</summary><pre>
NAME="Ubuntu"
VERSION="20.04.5 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.5 LTS"
VERSION_ID="20.04"
</pre></details>

j-wags:
## Description

In https://github.com/openforcefield/openff-toolkit/pull/1498 we introduced helpful logic to reindex the keys for any values in `offmol.properties['atom_map']` so that they correspond to the new ordering. It turns out that [BespokeFit ALSO did this](https://github.com/openforcefield/openff-bespokefit/blob/1b8763838c8070a4abee253bd72d66a40b3dacaf/openff/bespokefit/utilities/molecule.py#L96-L99), leading to a double-reindexing, which the tests caught and are warning us about. This can be fixed by removing the now-redundant logic from BespokeFit. 

## Todos
  - [ ] Resolve #237
  - [ ] Update releasehistory

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/utilities/molecule.py
+++ b/openff/bespokefit/utilities/molecule.py
@@ -92,12 +92,6 @@ def canonical_order_atoms(molecule: Molecule):
     atom_map = {i: rank for i, rank in enumerate(atom_order)}
 
     molecule = molecule.remap(atom_map, current_to_new=True)
-
-    if "atom_map" in molecule.properties:
-        molecule.properties["atom_map"] = {
-            atom_map[i]: j for i, j in molecule.properties["atom_map"].items()
-        }
-
     return molecule
 
 


mattwthompson:
Naively this looks like the fix for #234?

j-wags:
Hard to tell if this is a case of too much coffee or too little 🤦

jthorton (approved):
Great catch @j-wags this makes sense to me!

openff/bespokefit/utilities/molecule.py
```diff
@@ -92,12 +92,6 @@ def canonical_order_atoms(molecule: Molecule):
     atom_map = {i: rank for i, rank in enumerate(atom_order)}
 
     molecule = molecule.remap(atom_map, current_to_new=True)
-
-    if "atom_map" in molecule.properties:
-        molecule.properties["atom_map"] = {
-            atom_map[i]: j for i, j in molecule.properties["atom_map"].items()
-        }
-
     return molecule
```
Yoshanuikabundi, line 95: This fix should work for Toolkit 0.12.0, but will change behavior on 0.11.0 because the atom map won't be updated. I think the more robust fix is to make a copy of the atom map before remapping and then use that to produce the new mapping, something like:

```suggestion
    atom_map = {i: rank for i, rank in enumerate(atom_order)}
    original_atom_map = dict(molecule.properties["atom_map"])

    molecule = molecule.remap(atom_map, current_to_new=True)

    if "atom_map" in molecule.properties:
        molecule.properties["atom_map"] = {
            atom_map[i]: j for i, j in original_atom_map["atom_map"].items()
        }

    return molecule
```

Yoshanuikabundi (changes requested):
Thanks @j-wags! This is a good catch, but I think we need a slight tweak to support both Toolkit 0.11 and 0.12.

mattwthompson:
If both minor versions of the toolkit must be supported, should those be added as separate versions in CI?

j-wags:
I've added the suggested changes - I think this should be good for another look now!

Re: Testing against earlier versions - yeah, that'd be nice, but as you imply it's also be impractical. This solution seems elegant to me because it simply overwrites the changed behavior with exactly the desired logic. I'd trust this to work on the old version without adding CI testing. 

I'd have been slightly in favor of pinning `openff-toolkit>=0.12.1` in the package recipe, but I'm not the owner and this is also a fine resolution.

Yoshanuikabundi (approved):
Perfect, thanks!

I feel like supporting 0.11 and 0.12 just keeps our options open.

Yoshanuikabundi:
Hey @j-wags is there something holding this up? It's blocking testing on a number of other PRs so I think I'll merge it myself tomorrow if I don't hear back :)

j-wags:
Sorry for the delay - merged!

PR #236, opened 2023-03-06
[WIP] Try adding fast psi4 integration test

j-wags:


j-wags:
This PR should NOT be merged in its current form as it will obliterate the other testing - We'll want to decide how to integrate this into the testing suite (Josh M mentioned having this be a "slow nightly" test, which I think would be a good idea).

PR #239, opened 2023-03-15
QC-program keywords

### LINKED ISSUE #238: fitting failing due to too many open files
Hi,

I have bespokefit 0.1.3 running on ubuntu 20.04, installed via conda. I'm currently running in production mode and submitting via command line. I have not tried using the API yet. It works great on most molecules, however I've found a handful that die with the error:

[✓] fragmentation successful
[x] qc-generation failed

[null, {"type": "OSError", "message": "[Errno 24] Too many open files", "traceback": "Traceback (most recent call last):\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/celery/app/trace.py\", line 451, in trace_task\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/celery/app/trace.py\", line 734, in __protected_call__\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/openff/bespokefit/executor/services/qcgenerator/worker.py\",
 line 133, in compute_torsion_drive\n    input_schema, \"torsiondrive\", raise_error=True, local_options=_task_config()\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/openff/bespokefit/executor/services/qcgenerator/worker.py\",
 line 43, in _task_config\n    (psutil.virtual_memory().total / (1024**3))\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/psutil/__init__.py\", line 1968, in virtual_memory\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/psutil/_pslinux.py\", line 415, in virtual_memory\n  File
 \"~/miniconda3/envs/bespoke/lib/python3.9/site-packages/psutil/_common.py\", line 728, in open_binary\nOSError: [Errno
 24] Too many open files: '/proc/meminfo'\n"}]

I tried changing the allowed number of files to 4096 using ulimit but I still get the error. Googling around, it looks like this is a known issue with xtb writing out too many intermediate files, which they fixed with a verbosity setting. ([https://github.com/grimme-lab/xtb-python/issues/73](issue) and implemented in libxtb.py and interface.py in the xtb-python distribution) Is there a way to set the verbosity from the command line or do I need to dive into the API?

Thanks!

jthorton:
## Description
Fixes #238 by adding a program-based keyword generator until we have a general interface to accept keywords.

## Todos
  - [ ] TODO 1

## Questions
- [ ] should I also extend the [qc task](https://github.com/openforcefield/openff-bespokefit/blob/1b8763838c8070a4abee253bd72d66a40b3dacaf/openff/bespokefit/schema/tasks.py#L15) to accept keywords so the user can set them? Then the new function `_get_program_keywords` could add program-specific keywords needed from our experience that are missing from the provided dictionary.

## Status
- [ ] Ready to go


jthorton:
needs #235 for testing faliures

Yoshanuikabundi:
Thanks for looking into this @jthorton! Can I just make sure I understand what's going on here?

We expose QCSubmit's `QCSpec` class to users to define the calculation they want to make, but then when defining the tasks we want to run we only take the method, basis, and program and everything else is ignored. This happens [here](https://github.com/openforcefield/openff-bespokefit/blob/1b8763838c8070a4abee253bd72d66a40b3dacaf/openff/bespokefit/workflows/bespoke.py#L452C8-L467) (is `default_qc_specs` used anywhere else? My IDE couldn't find anything else).

We want to add the `'muted'` verbosity keyword to fix #238, so we're adding some code to add that keyword at the end to the specification of the QC job as an interim fix until something better can be worked out.

Have I got that right? If that's right, then I think this is great. I might also add a note to the big refactor issue that we should move away from using `QCSpec` here, since we don't support most of its settings?

jthorton:
Hi, @Yoshanuikabundi that's right! 

Our main reason for using `QCSpec` is the validation it provides on the combination of method, basis, program and implicit solvent but a few other fields are redundant like the spec name or description. Ideally, I think we would offer a base model in qcsubmit which still has this validation but with these redundant terms removed. 

The main features we will want here are validation of the method, basis and program combination along with the ability to pass keywords and possibly the validation of implicit solvent settings.

Yoshanuikabundi (approved):
This looks good! I'd like to add tests as we add new features, so I've written one in #244 - It's a PR to this PR. If you like it (after any tweaks you wanna make), could you merge it in here and make sure CI passes before merging this? Thanks @jthorton!

jthorton:
Thanks for adding the test. 

> do you know if its possible for the keyword to be set incorrectly in the input and then not get passed on to XTB without raising an error?

With most keywords, if it is misspelt it just won't be set by the wrapper around the program, and errors will only be raised if the value is incorrect in most cases. This is why we tried to build some validation into the `QCSpec` class to catch these before run time and would be another point in favour of more specialised program-specific keyword validation. 

With regard to testing this case, it would make sense to check that the output is empty from each single point calculation to confirm the keyword was correct and accepted however the result is currently striped by the worker [here](https://github.com/openforcefield/openff-bespokefit/blob/1b8763838c8070a4abee253bd72d66a40b3dacaf/openff/bespokefit/executor/services/qcgenerator/worker.py#L136-L139) to make the results objects smaller. The test does check that the keyword was inserted though which in this case is the best we can do I would say.

PR #256, opened 2023-04-28
Parallel torsiondrive optimisations

jthorton:
## Description
This PR tries to use the parallel torsiondrive optimization procedure via a threadpool.

The optimisations now run in testing where as they raised an error with a process pool which is positive however I still need to confirm that this is faster. I suspect in most cases this will only make a difference when using full QM. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/qcgenerator/qcengine.py
+++ b/openff/bespokefit/executor/services/qcgenerator/qcengine.py
@@ -1,4 +1,4 @@
-from multiprocessing import Pool
+import concurrent.futures
 from typing import Dict, List, Union
 
 from qcelemental.models import FailedOperation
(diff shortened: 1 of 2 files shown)


Yoshanuikabundi:
Hey @jthorton, sorry I'm only just seeing this now because iiiii had my notifications misconfigured :sweat_smile: 

I think I'm missing some context around what this is for - are we currently using this by default to parallelise torsion drives, or is this an experimental option? If the former, why don't we get errors in production? If the latter, could you show me a snippet of how to run this?

It would be really nice if we could submit these as celery jobs instead of having a pool within a pool - I think that would help avoid situations in which we're oversubscribing threads. But I'm not sure if that's possible with the current infrastructure and I don't know anything about celery so :shrug:

It makes sense to me that this could work - each thread could acquire the GIL, run until it launches the QC program, and then release the GIL while the QC runs to allow the next thread to set off its job. But I could also imagine it failing, for example by not releasing the GIL when the QC program is launched, or because the QC program is a child process (?). Did you get around to checking if this speeds things up?

Yoshanuikabundi:
Our `temporary_cd` function might be problematic with threads: https://github.com/openforcefield/openff-toolkit/issues/1642#event-9645791825

jthorton:
Thanks for the feedback @Yoshanuikabundi, I agree that I can see this not working as well and would prefer to use a process pool if possible as its the simplest way to split up the optimisations. The current issue is that when the workers are launched via the CLI they are already child processes and can't then start another pool. However, I think that #272 offers a way around this as the workers can be launched separately from the server and should be able to spwan a pool! This should speed things up a lot for users with a HPC setup! so ill come back to this once the PR is in.

> It would be really nice if we could submit these as celery jobs

This would be great and probably the most efficient use of resources however it adds a lot of complexity to keep track of running optimisations and extend the qcgeneration service to collect and spawn new optimisations, at this point we would be recreating what QCFractal can do and since we already have an interface to pull results from there we should point users to that sort of setup in the docs if they need that kind of throughput? For now, I would like to see how far we can get with the simple approach of using a pool and then look at the more complicated options if we have to.

PR #260, opened 2023-05-04
Fix #259

### LINKED ISSUE #259: Circular import when importing `BespokeWorkflowFactory`.
**Description**

The new `tempcd` module imports `from openff.bespokefit.executor.services import current_settings`. This can cause a circular import. A workaround is to call `import openff.bespokefit.executor.services._settings`, which pre-loads the required module in the right order.

**Reproduction**

Broke:

```python
from openff.bespokefit.workflows import BespokeWorkflowFactory
```

Note that, in iPython or presumably Jupyter, entering this line via tab completion *will* work (ie, will not raise an exception), because the offending module will get pre-loaded when the kernel searches for completions. Which leads us to a workaround:

Woke:

```python
import openff.bespokefit.executor.services._settings
from openff.bespokefit.workflows import BespokeWorkflowFactory
```

**Fixing**

We should rush out a patch to move the settings to the utils module - there's no reason for it to be so deep in the module tree. This'll prevent the entire executor module from getting loaded just to access the settings. We can leave the existing API point as a re-export until the next breaking release.

Yoshanuikabundi:
Fixes #259 by moving the `_settings` module to utilities.

@j-wags Is it OK if I release 0.2.2 on my next working day to get this out there? 

## Status
- [x] Ready to go


mattwthompson:
This fixes the issue for me locally

j-wags (approved):
Yup, approved (though you don't need my approval :-) ) and feel free to make a release whenever you deem fit.

PR #299, opened 2023-11-14
Update installation docs

mattwthompson:
## Checklist
- [x] Bump Python version to 3.10
- [x] Force `qcportal <0.50`
- [x] Remove references to `-c psi4`
- [x] Remove references to `-c defaults`
- [x] Use [blessed](https://psicode.org/installs/v182/) Psi4 installation pathway, which currently uses `-c conda-forge/label/libint_dev`
- [x] Update call to `python setup.py`

## Status
- [ ] Ready to go

PR #300, opened 2023-11-14
Fix rendering

mattwthompson:
I neglected to look at the docs rendered in #299


mattwthompson:
I don't see a docs build on the PR status checks?

image

mattwthompson:
In any case, the build after #299 is pretty simply and obviously broken: 
image


I'm just going to fire this off and hope for the best.

Issue #305, opened 2023-12-04
Automate running Psi4-based fits using default installation instructions


mattwthompson:
**Description**

I don't think any of the tests run a full fit using the default QC spec (i.e. not using XTB as a shortcut). I could be wrong.

These fits are slow but important to ensure reliability - if something breaks, we ought to know about it prior a user report.

mattwthompson:
Quick thought in passing - the issues leading us to work on #308 arise from one of Psi4's dependencies not being installed, so safeguarding against that particular failure only requires enough walltime for the QC generation step to _start_, not actually complete.

jthorton:
It might be overly complicated but would a test which runs a fit but mocks the QC generation stage work? We could have it try and compute a dft single point using our default spec in psi4 (this uses a D3 correction) and if that works return some already computed DFT torsion profile and the fit can continue. The test would fail if the single point failed and that should catch install issues.

mattwthompson:
If a single-point energy would capture most "oops something is wrong in the Psi4 installation" errors - yeah I think that could work nicely.

In my limited experience, user issues happen either before QC generation (nothing is installed correctly) or in the first 0.1% of the QC generation step (current or past issues with Psi4 installation), so mocking the results of QC generation could be a nice balance of reliability testing and test duration.

j-wags:
Unrelated but the call stack reminds me of the star wars [assassins-hiring-assassins video](https://www.youtube.com/watch?v=S5YWbqKVpwE).

PR #308, opened 2023-12-05
Fix install instructions

j-wags:


docs/getting-started/installation.md
```diff
@@ -72,7 +72,7 @@ mamba create -n bespokefit-env -c conda-forge/label/libint_dev -c conda-forge py
 or with OpenEye Toolkits:
 
 ```shell
-mamba create -n bespokefit-env -c conda-forge-label/libint_dev -c conda-forge -c openeye python=3.10 openff-bespokefit "qcportal <0.50" psi4 openeye-toolkits
+mamba create -n bespokefit-env -c conda-forge/label/libint_dev -c conda-forge -c openeye python=3.10 openff-bespokefit "qcportal <0.50" psi4 openeye-toolkits
```
mattwthompson: Oof ... my bad.

This should also have `-c psi4 ... dftd3`

mattwthompson (approved):
Mixing channels really makes me nervous but there's no way around it here - unless one can use the `conda-forge::psi4` trick on the command line

Safeguards might involve `"psi4=1.8"` or #305

docs/getting-started/installation.md
j-wags: I think I caught that one as you were typing the comment :-) Thanks for another set of eyes.

mattwthompson:
I might ask to hold off on merging this, though. I've asked the Psi4 folks about the change between `psi4::psi4` and `conda-forge::psi4` since I find it confusing. We probably will want this change, but it's possible there's a different option that's recommended.

mattwthompson:
There's `conda-forge::dftd3-python` (and `conda-forge::dftd4-python`, and `conda-forge::gcp-correction` if those are useful) which is what we're supposed to use instead of anything from the `psi4` channel. We might want to do a quick test fit to ensure it actually works. I'm not totally sure why the packaging changed in the channel migration, but this is an iceberg of complexity I'm okay not fully understanding.

mattwthompson:
Here's a conversation between Lori and I and earlier today from the #building_packging channel in the Psi4 slack server (I won't ping her here, since this is a tsunami of context without a discrete ask, but if our troubleshooting doesn't work we can formulate another question to bring back to her.)

Me:
> How does `psi4::dftd3` differ from similarly-named packages on `conda-forge`? I notice it’s one of the things that used to be pulled down with `psi4::psi4` but isn’t there with `conda-forge::psi4`, so we’ll need `-c psi4` which otherwise would not be present. From my searches so far, it seems like it’s [used](https://github.com/psi4/psi4/pull/2791) (?) but only if it’s found (i.e. not installed by default) by the `conda-forge::dftd3-python` package. Not a complaint, just trying to better understand things. (My naivety of the quantum chemistry and the various DFT things out there int he world doesn’t exactly help me here)

LB:
> only in older or compatibility-check specs should `psi4::dftd3` be pulled down. https://github.com/psi4/psi4/pull/2791 basically, v1.7 allowed the newer c-f ones, v1.8 started preferring them if both available, and v.1.9 stops supporting classic, though it’s not disabling them.

Me:
> So is `conda-forge::dftd3-python` the same thing (or a newer version thereof) as `psi4::dftd3`? And should I have users pull down `conda-forge::dftd3-python`?

LB:
> well, `conda-forge::dftd4-python` note d4, not d3, is the same thing (or a newer version thereof) as `psi4::dftd4`. `conda-forge::dftd3-python` is a complete rewrite and new API interface to the code that is `psi4::dftd3` with a parsed interface. But yes, have users pull down `c-f::dftd3-python` `c-f::dftd4-python` `c-f::gcp-correction` rather than anything from psi4 channel. similar info in:
* https://github.com/psi4/psi4/blob/master/codedeps.yaml#L306
* running `conda/psi4-path-advisor.py env`
* viewing env spec (generated from codedeps by path-advisor) for any Eco Linux GHA run: https://github.com/psi4/psi4/actions/runs/7106466510/job/19345951510#step:6:49


----

From this I think `conda-forge::dftd4-python` is worth a shot. I interpret above to mean that one of these three or so packages needs to be present for something important to Psi4, and it magically works with whatever's available modulo some details of their upgrade schedule. (To be honest I don't understand how or where BespokeFit actually interfaces with Psi4, much less how various `.*dftd[3-4].*` packages interact with it). My general confusion of why some package did not persist in the `conda-forge` migration I think is answered by the big table in the linked PR

j-wags:
Thanks for the context, that's really helpful. 

I wouldn't expect dftd4 to be a replacement for- or superset of- dftd3's functionality, but that's an educated guess with no context so it's very much worth checking. 

I think a more evolved version of the table from the PR made it into the [psi4 docs](https://psicode.org/psi4manual/master/dftd3.html). It helpfully puts version numbers to the compatibility ranges.

I can only get a default-qc-spec bespokefit job to work locally (on an M1 mac) with `dftd3-python`, but Bill could only get it working on his linux cluster with `psi4::dftd3`. If `psi4::dftd3` is present on my mac (even if `dftd3-python` and `simple-dftd3` are ALSO there), the jobs fail for me. For Bill, the jobs failed when only `dftd3-python` and `simple-dftd3` were installed, and only started working once `psi4::dftd3` was installed. 

`gcp-correction` is new to me. I don't _think_ it's relevant to us, since I don't see it in Pavan's or my envs, though that could be another avenue to look down. 

I need to experiment more with configurations, will try spinning up a linux/intel machine tomorrow to test. Bill said he'd be willing to experiment with other env creation commands if we need him to. Worst case is it's some system library issue on Bill's cluster and we will have to provide a note like "if your system is old you may need to downpin psi4".

PR #312, opened 2023-12-15
Use Psi4 1.9 in documentation

mattwthompson:
## Description
This PR drops the `conda-forge/label/libint_dev` label which is no longer needed.

## Status
- [ ] Ready to go

### DIFF
--- a/docs/getting-started/installation.md
+++ b/docs/getting-started/installation.md
@@ -51,7 +51,7 @@ mamba install -c openeye openeye-toolkits
 recommended to be installed unless you intend to train against data generated using a surrogate such as ANI:
 
 ```shell
-mamba install -c conda-forge/label/libint_dev -c conda-forge psi4 dftd3-python
+mamba install -c conda-forge psi4 dftd3-python
 ```
 
 [Psi4]: https://psicode.org/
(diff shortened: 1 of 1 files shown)


mattwthompson:
```shell
$ git checkout psi4-1.9 && grep -r libint . | wc -l             14:27:17  ☁  5bb7767 ☂ ✭
Previous HEAD position was 5bb7767 Bump actions/upload-artifact from 3 to 4 (#311)
Switched to branch 'psi4-1.9'
       0

j-wags:
@Yoshanuikabundi reminds me that, just before the holidays, we were having package compatibility trouble between conda-forge's psi4 (using conda-forge libint) and NAGL (which uses conda-forge's pytorch). He'll try it again and will report back if the incompatibility is still an issue, and whether there's a issue we can follow for it's resolution. 

Either way, that shouldn't block merging this, so @Yoshanuikabundi should go ahead and review. It'll be very rare for users to need an env with both NAGL and psi4 (mostly that will be our science team, and we can help them with that). And hopefully when NAGL does have a full rollout as an AM1BCC provider in a few months the package incompatibility will have shaken out.

j-wags (approved):
Tested locally and this worked great. And docs.openforcefield.org directs people to the `latest` bespokefit docs so this should go live without needing a release!

PR #317, opened 2024-01-29
Update test environments

mattwthompson:
## Description
- [x] Ensure compatible versions of OpenFF and QCArchive software are brought in
- [x] Remove obsolete shims
- [x] Remove unused `codecov` package

## Status
- [x] Ready to go


j-wags:
Ahhh, good catch on this. I'd checked the that the feedstock pins were good, but not the CI envs. Thanks!

PR #331, opened 2024-03-11
Do not cite Rosemary

mattwthompson:


j-wags (approved):
Wow, I can't imagine how much low-level confusion this has caused over the years. Great catch!

PR #280, opened 2023-08-04
Use new QCFractal

mattwthompson:

PR #332, opened 2024-03-22
Also test on Python 3.11

mattwthompson:
We ought to test on the most recent versions we can. Users are likely using other versions: https://prefix.dev/channels/conda-forge/packages/openff-bespokefit

It might be worth revisiting after #280 is merged in


mattwthompson:
I'll be merging this in when it's green again, dictatorially so without dictator authority

j-wags (approved):

Issue #340, opened 2024-04-11
Make docker container 


hmacdope:
It would be great if you could deploy bespokefit with a docker container.

hmacdope:
We can also publish docker packages with `ghcr.io`

j-wags:
What's the motivation for deploying on docker? We have to provide documentation and user support for anything we release so I want to make sure this is something that provides our partners value and that we have the time and expertise to maintain. This seems simpler than an entirely new deployment pathway (since it's still just conda under the hood), but every deployment pathway that we have to maintain is an added cost. 

We have a [dockerhub account](https://hub.docker.com/orgs/openforcefield/repositories) that I could give you access to for hosting the images if we go ahead with this. Would the docker image have psi4 and/or xtb (and/or other engines)? And what would the automated testing for this look like?

hmacdope:
Fair points @j-wags, I should have explained motivation a bit better. 

A Dockerfile based container is a very easy way to spool up a service on a host, with all the dependencies and configuration already taken care of, alongside logs and easy status tracking.

 This is most useful in a non HPC based environment, where the server is intended to not be ephemeral, but rather exist in an *always on* mode. This is very common outside of academia and on cloud instances. Docker is pretty much the standard for service based architectures to package their work for easy deployment. Running a `python` process in tmux is probably too fragile for a more heavyweight deployment.  

There are two separate issues here that we should try not to conflate. 

1) Having a `Dockerfile`. This  provides people the ability to start `bespokefit-server` as a docker service by building it themselves. 

e.g start a working bespokefit-server instance in two commands (you could do similar for the `executor` service if you wanted.)

```bash
git clone git@github.com:openforcefield/openff-bespokefit.git
docker-compose up -d
# profit 
```

Possible value adds
 * Always-works portable deployment on any hardware
 * Opens up new types of deployments that need more horizontal scalability or error tolerance 
 * Build isolation from the host, better for stability and security
 * Deployable in a microservices based model, where bespokefit has to fit into a chain of services that communicate with each other

A lot of these are probably more relevant to industry than in an academic lab, which is why I wanted to pass this on as an industry adjacent perspective. 

2) Deploying a `Docker` based container to `dockerhub` or `ghcr.io`, container registries that house pre-built containers. I prefer `ghcr.io` as already integrated with github, but `dockerhub` also very easy to integrate. 

Possible value adds
 * Consistent builds you can point people to, no need to debug people's build environments, they can just use official one.
[...]

PR #342, opened 2024-04-14
Add docker package build 

hmacdope:
Branches off #340, so that will have to go in first. 

Adds a CI workflow that publishes a docker container to `ghcr.io` on push to main. 

Just a suggestion, if yall don't want or need this feel free to close.


hmacdope:
Alternatively, could build package on `release` trigger rather than on `main`

PR #351, opened 2024-06-12
BespokeFit client

jthorton:
## Description
This PR introduces a simple bespokefit client which splits out the logic of communicating with the executor. This should make it easier to submit and retrieve calculations from an executor on another machine using the python API similarly to Alchemiscale or QCArchive. To enable this we also expose the `BEFLOW_GATEWAY_ADDRESS` which can be used to provide the location of an executor on another machine which is currently not possible. 

We also introduce some basic authentication which checks that the value of the new setting `BEFLOW_API_TOKEN` matches between the server and the client value before responding to requests. 

These features should make it easier to provide a light weight bespokefit-client package in future for applications which only need to interface with a bespokefit executor. 

## Todos
  - [x] add authentication and testing
  - [x] add bespokefit client and testing
  - [x] update all CLI commands
  - [x] update docs using the old interface

## Questions
- [ ] ~~should we keep the old API points and raise a deprecation warning if they are used?~~ 
- [x] Patch the methods on the executor to go via the client

## Status
- [x] Ready to go

### DIFF
--- a/openff/bespokefit/_tests/executor/test_executor.py
+++ b/openff/bespokefit/_tests/executor/test_executor.py
@@ -10,7 +10,6 @@ from openff.bespokefit.executor import (
     BespokeExecutor,
     BespokeExecutorOutput,
     BespokeExecutorStageOutput,
-    wait_until_complete,
 )
 from openff.bespokefit.executor.services import current_settings
 from openff.bespokefit.executor.services.coordinator.models import (
@@ -411,9 +410,9 @@ def test_wait_for_stage(status):
     assert response.results is None
 
 
-def test_wait_until_complete():
-    with mock_get_response("success") as mock_response:
-        response = wait_until_complete(mock_response.id, frequency=0.1)
-
-    assert isinstance(response, BespokeExecutorOutput)
-    assert response.stages[0].status == "success"
+# def test_wait_until_complete():
+#     with mock_get_response("success") as mock_response:
+#         response = wait_until_complete(mock_response.id, frequency=0.1)
+#
+#     assert isinstance(response, BespokeExecutorOutput)
+#     assert response.stages[0].status == "success"
--- a/openff/bespokefit/_tests/executor/test_executor.py
+++ b/openff/bespokefit/_tests/executor/test_executor.py
@@ -10,6 +10,7 @@ from openff.bespokefit.executor import (
     BespokeExecutor,
     BespokeExecutorOutput,
     BespokeExecutorStageOutput,
+    BespokeFitClient,
 )
 from openff.bespokefit.executor.services import current_settings
 from openff.bespokefit.executor.services.coordinator.models import (
@@ -277,7 +278,7 @@ class TestBespokeExecutor:
 
     def test_submit(self, bespoke_optimization_schema):
         settings = current_settings()
-
+        client = BespokeFitClient(settings=settings)
         expected = CoordinatorPOSTResponse(id="mock-id", self="")
 
         with requests_mock.Mocker() as m:
@@ -291,12 +292,13 @@ class TestBespokeExecutor:
                 text=expected.json(),
             )
 
-            result_id = BespokeExecutor.submit(bespoke_optimization_schema)
+            result_id = client.submit_optimization(bespoke_optimization_schema)
             assert result_id == expected.id
 
     def test_retrieve(self, bespoke_optimization_schema):
         with mock_get_response():
-            output = BespokeExecutor.retrieve("mock-id")
+            client = BespokeFitClient(settings=current_settings())
+            output = client.get_optimization("mock-id")
             assert output.status == "running"
 
     def test_enter_exit(self, tmpdir):
@@ -325,7 +327,6 @@ class TestBespokeExecutor:
 
 
 def test_query_coordinator():
-    from openff.bespokefit.executor.executor import _query_coordinator
 
     mock_response = CoordinatorGETResponse(
         id="mock-id",
@@ -343,6 +344,7 @@ def test_query_coordinator():
         return mock_response.json(by_alias=True)
 
     settings = current_settings()
+    client = BespokeFitClient(settings=settings)
 
     mock_href = (
         f"http://127.0.0.1:"
@@ -354,14 +356,13 @@ def test_query_coordinator():
 
     with requests_mock.Mocker() as m:
         m.get(mock_href, text=mock_callback)
-        response = _query_coordinator(mock_href)
+        response = client._query_coordinator(mock_href)
 
     assert response.json() == mock_response.json()
 
 
 @pytest.mark.parametrize("status", ["success", "errored"])
 def test_wait_for_stage(status):
-    from openff.bespokefit.executor.executor import _wait_for_stage
 
     mock_response = CoordinatorGETResponse(
         id="mock-id",
@@ -388,6 +389,7 @@ def test_wait_for_stage(status):
         return response_json
 
     settings = current_settings()
+    client = BespokeFitClient(settings=settings)
 
     mock_href = (
         f"http://127.0.0.1:"
@@ -400,7 +402,9 @@ def test_wait_for_stage(status):
     with requests_mock.Mocker() as m:
         m.get(mock_href, text=mock_callback)
 
-        response = _wait_for_stage(mock_href, "fragmentation", frequency=0.01)
+        response = client._wait_for_stage(
+            mock_response.id, "fragmentation", frequency=0.01
+        )
 
     assert response is not None
     assert n_requests == 2
--- a/openff/bespokefit/_tests/executor/test_executor.py
+++ b/openff/bespokefit/_tests/executor/test_executor.py
@@ -414,9 +414,11 @@ def test_wait_for_stage(status):
     assert response.results is None
 
 
-# def test_wait_until_complete():
-#     with mock_get_response("success") as mock_response:
-#         response = wait_until_complete(mock_response.id, frequency=0.1)
-#
-#     assert isinstance(response, BespokeExecutorOutput)
-#     assert response.stages[0].status == "success"
+def test_wait_until_complete():
+    settings = current_settings()
+    client = BespokeFitClient(settings=settings)
+    with mock_get_response("success") as mock_response:
+        response = client.wait_until_complete(mock_response.id, frequency=0.1)
+
+    assert isinstance(response, BespokeExecutorOutput)
+    assert response.stages[0].status == "success"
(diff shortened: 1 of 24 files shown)


jthorton:
Hi @j-wags , @mattwthompson , @Yoshanuikabundi I wasn't sure who to ask for a review here so feel free to ignore if I got the wrong person! The main part I need feedback on is if I should keep the old methods on the executor and add deprecation warnings to them or if I can get away with removing them and having this be a major version change, happy to do either though. Also, any feedback generally would be great!

j-wags:
I'll give this a review today, mostly with an eye on the API change/deprecation questions.

openff/bespokefit/executor/executor.py
```diff
@@ -349,114 +216,9 @@ def _stop(self):
         if self._remove_directory:
             shutil.rmtree(self._directory, ignore_errors=True)
 
-    @staticmethod
-    def submit(input_schema: BespokeOptimizationSchema) -> str:
```
j-wags, line 353: It'd be very bad to remove an API point that's used in the quickstart guide with no deprecation period. This would drop a fire in the lap of everyone who uses BespokeFit in a workflow and lose a lot of user trust. I'd recommend either:
* Adding a deprecation warning for ~3+ months 
* Not deprecating this at all, and just long-term redirecting to `client.submit_optimization`

openff/bespokefit/executor/executor.py
```diff
@@ -349,114 +216,9 @@ def _stop(self):
...
-        request = requests.post(
-            _coordinator_endpoint(),
-            data=CoordinatorPOSTBody(input_schema=input_schema).json(),
-        )
-        request.raise_for_status()
-
-        return CoordinatorPOSTResponse.parse_raw(request.text).id
-
-    @staticmethod
-    def retrieve(optimization_id: str) -> BespokeExecutorOutput:
```
j-wags, line 371: Ditto - this also appears in the quickstart

openff/bespokefit/executor/services/coordinator/stages.py
```diff
@@ -144,7 +145,8 @@ async def _update(self):
                 f"http://127.0.0.1:"
                 f"{settings.BEFLOW_GATEWAY_PORT}"
                 f"{settings.BEFLOW_API_V1_STR}/"
-                f"{settings.BEFLOW_FRAGMENTER_PREFIX}/{self.id}"
+                f"{settings.BEFLOW_FRAGMENTER_PREFIX}/{self.id}",
+                headers={"bespokefit-token": settings.BEFLOW_API_TOKEN},
```
j-wags, line 149: (I'm not in any way familiar with internet security, just wanted to raise this because it seemed smelly) Is it safe to transmit API tokens using vanilla `http` or is this a use case for `https`?

j-wags (commented):
For the docs update, it'd be good to ensure that they reflect best practices - so examples/docs pages should be updated to avoid things that are removed/emit deprecation warnings.

openff/bespokefit/executor/services/coordinator/stages.py
jthorton: Ahh good catch this is where the server is connecting to itself but yes probably still good to use `https`.

jthorton: But for some reason this breaks the server 
```
"SSLError: [SSL] record layer failure (_ssl.c:1006)"
```

openff/bespokefit/executor/executor.py
jthorton: For now, I went with not deprecating it and just redirecting the methods to use the client.

openff/bespokefit/_tests/executor/test_executor.py
```diff
@@ -411,9 +414,9 @@ def mock_callback(request, context):
...
-        response = wait_until_complete(mock_response.id, frequency=0.1)
-
-    assert isinstance(response, BespokeExecutorOutput)
-    assert response.stages[0].status == "success"
+# def test_wait_until_complete():
+#     with mock_get_response("success") as mock_response:
+#         response = wait_until_complete(mock_response.id, frequency=0.1)
+#
+#     assert isinstance(response, BespokeExecutorOutput)
+#     assert response.stages[0].status == "success"
```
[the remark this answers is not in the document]
jthorton: Ah no good find!

jthorton:
Thanks @mattwthompson and @j-wags! I think I have covered everything now but it would be great if you could give the docs a quick look over to make sure I didn't miss an old use of the executor!

openff/bespokefit/executor/client.py
```diff
@@ -0,0 +1,292 @@
...
+        self,
+        settings: Settings,
+        retries: int = 3,
+        backoff_factor: float = 0.5,
+    ):
+        self._session = requests.Session()
+        self._session.headers.update({"bespokefit-token": settings.BEFLOW_API_TOKEN})
+        retry = Retry(connect=retries, backoff_factor=backoff_factor)
+        adapter = HTTPAdapter(max_retries=retry)
+        self._session.mount("http://", adapter)
```
j-wags, line 146: @dotsdl if you have a moment - I know next to nothing about internet security, but this smells a little funny to me. If we're adding API tokens to bespokefit, is it safe to put them in headers over http, or should we be using https?

j-wags (commented):
Thanks for the changes @jthorton. Everything looks good, except I'm still concerned about sending the tokens (or any other sensitive data) over raw http. I did some searching on my own and [asked ChatGPT](https://chatgpt.com/share/3669b9fd-a2ac-4f59-9856-f5d7f73aca3c), and it seems like passing sensitive data between machines does require https. I don't have the knowledge to recognize whether this is happening in a context where we don't need https, so I've asked Dotson to weigh in.

openff/bespokefit/executor/client.py
jthorton: Good point I would have expected that as well, I checked Alchemiscale and it turns out you can use something like [traefik](https://github.com/traefik/traefik) to automatically do this for you without having to implement it in the package. You can see that this is part of the deployment docker file in [alchemiscale](https://github.com/openforcefield/alchemiscale/blob/938690746692dbfa2d4b97c2423495ccb26a6b2b/docker/alchemiscale-server/docker-compose.yml#L172).

dotsdl: @bennybp what are you using for handling TLS on your QCFractal deployments? We we `traefik` to handle this for `alchemiscale`, but curious what you've been using.

bennybp: Same here. All the instances (and other projects) that are on that server run behind traefik, which handles all the routing and certificates. QCFractal doesn't have direct support for SSL itself (at least I don't see it as on option for `waitress`)

dotsdl:
I think it's generally wise to offload handling of TLS to a TLS termination proxy, such as `traefik`. This can also handle certificate renewals automatically when using e.g. Let's Encrypt as a certificate authority.

Since `openff-bespokefit` uses `fastapi` for its HTTP APIs, this page gives a great explanation for how these pieces fit together: https://fastapi.tiangolo.com/deployment/https/

Do we have docs on how to deploy `openff-bespokefit`'s server components, especially for a long-lived instance accessible across the network/internet? What may make sense is to offer a `docker-compose`-based deployment path like we do [for alchemiscale](https://github.com/openforcefield/alchemiscale/tree/main/docker/alchemiscale-server), since this includes deployment and configuration of `traefik` all in a single deployable config.

dotsdl:
That said, if a user deploys `openff-bespokefit`s server components to a host inside of a closed network (not reachable from the internet), you can get away with not using TLS for HTTPS communication and instead use only raw HTTP.

So, this comes down to decisions by an individual user/deployer as to how they plan to expose their `openff-bespokefit` instance.

jthorton:
Thanks @bennybp and @dotsdl for the response this definitely helps and hopefully clears things up for @j-wags. 

>What may make sense is to offer a docker-compose-based deployment path like we do [for alchemiscale](https://github.com/openforcefield/alchemiscale/tree/main/docker/alchemiscale-server), since this includes deployment and configuration of traefik all in a single deployable config.

Yes this is a great idea and something I want to add and started in #342 , but as you point out most of the time users won't need this as they can run it on a local machine, but hopefully we can document how to do this on AWS as we have with the ASAP deployment.

j-wags:
Yes, thank you both for the clarification. I'm fine with the client functionality being added, but I don't think the token field makes sense if the two situations where it's being used are:

1. when the client and server are on the same machine/a closed network, and no authentication is needed in the first place
2. when the client and server are communicating through the open internet, where third parties can see the token in plaintext

So I'm happy to consider this PR without any use of tokens and making no security guarantees (and a documented warning that using this over the open internet is not secure). Or I'm happy to have tokens with https-by-default to enable secure communication. But I don't want to put this out with seemingly-secure tokens that aren't actually secure.

mattwthompson:
Who owns this project now? @jthorton is currently listed as the contact https://github.com/openforcefield/status?tab=readme-ov-file#openff-maintainer-dashboard

j-wags:
Josh is indeed currently the owner of Bespokefit and could unilaterally merge this if he wanted. However this (to me) is a serious enough issue that I'd bring it up with the lead team if this went ahead in its current state. 

The context around this PR is that ASAP is building software pipelines to discover new drugs, and they must implement data security to avoid massive IP/legal trouble if the data leaks. Using structurally insecure token authentication presents a very large risk for them, and has the possibility of catching us in the blast radius if something goes wrong. And this is before the issue of us pushing insecure token authentication to OUR partners as well.

jthorton:
>Yes, thank you both for the clarification. I'm fine with the client functionality being added, but I don't think the token field makes sense if the two situations where it's being used are:

Yeah I could see users doing option 2 rather than option 3 which is to deploy behind treafik and thinking that the server is secure when its not, but I am not sure if its possible for us to automatically enforce/ enable https by default without adding a lot of complexity. I think the best thing we can do is to provide the docker file with treafik and detailed deployment docs outlining how this works and cases when it is secure similar to alchemiscale. 

We could also have warnings emitted if the user sets a authentication token and tries to connect to a http address rather than https and warn that its not secure?

j-wags:
To recap our discussion in the Newcastle/bespoke meeting today - Josh's proposal to emit a warning when a client connects to anything except `localhost` or `0.0.0.0` using http works for me. 

We didn't talk about steps forward for the dockerfile - I'd prefer that go into a separate PR given that the warning-on-http plan gets this to a mergeable state, and we'll need to do some thinking about the devops/testing/documentation of a docker deployment pathway.

jthorton:
Thanks @j-wags I have added the warning along with a test please feel free to suggest a changing to the wording of the warning I wasn't quite sure of how urgent to make the warning!

j-wags (approved):
Fantastic. Thanks @jthorton. This is good to merge, and I'm thinking it's a significant enough feature that it should be release 0.4.0. Any notes to add about reverse compatibility, or should previously working commands still work? Please update the releasenotes to list any gotchas here, but if that's not a concern this is good to merge.  

I'll assume that I should cut the release once this is merged but let me know if it should wait for any other PRs.

jthorton:
Thanks @j-wags, there should be no changes to past commands or scripts as we export all of the old functions still but run them through the new client feature and yes a 0.4.0 version would be great to mark this feature!

PR #357, opened 2024-09-05
[DNM] Test on pydantic 2

j-wags:


j-wags:
Looks like lots of failures here - Worth revisiting after Interchange 0.4 is out and using native pydantic v2 objects.

devtools/conda-envs/no_openeye.yaml
```diff
@@ -1,6 +1,7 @@
 name: bespokefit-no-openeye
 
 channels:
+  - conda-forge/label/openff-interchange_rc
```
mattwthompson, line 4: This isn't needed for Pydantic 2 compatibility

j-wags: Gotcha, rerunning tests without it

mattwthompson:
After isolating Pydantic as the change, some surprising things are happening: https://github.com/openforcefield/openff-bespokefit/pull/358

PR #358, opened 2024-09-09
Always use Pydantic v1 API, test using v1 and v2

mattwthompson:


mattwthompson:
The failures here are genuine, but could be due to hard-coded parts of the tests. I can't tell at the moment, but note that the traceback includes only files from upstreams.

SimonBoothroyd:
seems like this could related to https://github.com/fastapi/fastapi/issues/10360?

mattwthompson:
Could be - I haven't used the V1 API backdoor in a project that also uses fastAPI and it wouldn't surprise me if that's not possible.

I don't see any reason why this couldn't use V2, but it'd be some work

hmacdope:
We are in the process of migrating and are at the bespokefit part. I can try and help with migration if it would assist @mattwthompson @SimonBoothroyd

mattwthompson:
cc @j-wags

PR #360, opened 2024-09-19
try latest qcf

j-wags:
https://github.com/openforcefield/openff-bespokefit/pull/337
https://github.com/openforcefield/openff-bespokefit/issue/336


jthorton (approved):
Not sure why 3.10 and 3.11 testes are pulling in different versions of qcportal but they are all passing so this should be fine!

PR #361, opened 2024-10-01
Pin micromamba 1.5.10-0

Yoshanuikabundi:
## Description
This should work around mysterious CI failures relating to Micromamba - see https://github.com/mamba-org/micromamba-releases/issues/58

## Status
- [x] Ready to go


mikemhenry:
I think this has been fixed upstream now

Issue #298, opened 2023-11-14
Update installation docs


mattwthompson:
**Description**

The installation docs are a little out of date: https://docs.openforcefield.org/projects/bespokefit/en/stable/getting-started/installation.html

They include
* Psi4 pulling from `defaults`
* Psi4 not looking into `https://psicode.org/installs/v182/` which it currently recommends
  * https://psicode.org/installs/v182/
* Possibly pulling in `openff-qcsubmit >=0.50` or `qcportal >=0.50`

mattwthompson:
https://github.com/conda-forge/openff-bespokefit-feedstock/issues/19

mattwthompson:
In #299 I've updated them to reflect the current state of needing `qcportal <0.50`; this should not last long either with the above repodata patch merged or an upcoming release using `qcportal >=0.50`. But the live docs should be accurate

PR #366, opened 2024-10-10
Update install steps 

mikemhenry:
Doing `"qcportal <0.50"` will pull down version `0.2.3` of bespokefit


mattwthompson (approved):
Thanks for reporting this - this paper clip is super old and removing it is long overdue!

I think this will address what's left of #298

PR #384, opened 2025-02-26
Update QCSubmit cache with new `GeometricProcedure

### LINKED ISSUE #383: test_cache.py::test_update_from_qcsubmit is failing
For example, see https://github.com/openforcefield/openff-bespokefit/actions/runs/13510531841/job/37749688855

```
=================================== FAILURES ===================================
__________________________ test_update_from_qcsubmit ___________________________

redis_connection = <redis.client.Redis(<redis.connection.ConnectionPool(<redis.connection.Connection(host=localhost,port=5678,db=0)>)>)>

    def test_update_from_qcsubmit(redis_connection):
        """
        Test adding a mocked result to a local redis instance.
        """
    
        console = rich.get_console()
        qcsubmit_result = TorsionDriveResultCollection.parse_file(
            get_data_file_path(
                os.path.join("test", "schemas", "torsion_collection.json"),
                package_name="openff.bespokefit",
            )
        )
    
        with console.capture() as capture:
            _update_from_qcsubmit_result(
                console=console,
                qcsubmit_results=qcsubmit_result,
                redis_connection=redis_connection,
            )
    
        # The terminal output often contains colors that aren't always displayed the same;
        # these patterns miss lots of fancy stuff from rich
        for line in [
            " local results built",
            " results cached",
        ]:
            assert line in capture.get()
    
        # find the result in redis
        # key is generated by querying the redis_connection object as below, and
        # .hkeys appears to always return a length-1 list of bytes. The result is
        # hardcoded here, generated Feburary 2024
        # key = redis_connection.hkeys(name="qcgenerator:task-ids")[0]
        task_id = redis_connection.hget(
            name="qcgenerator:task-ids",
            key="2b9f751eb62f9e6af02461f8853ffaa1e9beaaccac71ce8eaec8a9ac574d94320438c33a2947a49a7cd6d4af7249d3936fccde8ebc4bcc9c7b3d61d7b0478388",
        )
        if task_id is None:
            # same as above, but key captured in an environment without OpenEye licensed
            task_id = redis_connection.hget(
                name="qcgenerator:task-ids",
                key="1207a547546633062091443e2ac3105735ab98803e9c027786ea0a66049909c7508d64755db29b5193c13887e37203b2901c73a15720e866ee100e4025007e6e",
            )
    
>       assert redis_connection.hget("qcgenerator:types", task_id) == b"torsion1d"

openff/bespokefit/_tests/cli/test_cache.py:153: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/commands/core.py:4970: in hget
    return self.execute_command("HGET", name, key, keys=[name])
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/client.py:559: in execute_command
    return self._execute_command(*args, **options)
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/client.py:567: in _execute_command
    return conn.retry.call_with_retry(
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/retry.py:62: in call_with_retry
    return do()
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/client.py:568: in <lambda>
    lambda: self._send_command_parse_response(
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/client.py:541: in _send_command_parse_response
    conn.send_command(*args, **options)
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/connection.py:557: in send_command
    self._command_packer.pack(*args),
../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/connection.py:107: in pack
    for arg in map(self.encode, args):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <redis._parsers.encoders.Encoder object at 0x7f5b1da959c0>, value = None

    def encode(self, value):
        "Return a bytestring or bytes-like representation of the value"
        if isinstance(value, (bytes, memoryview)):
            return value
        elif isinstance(value, bool):
            # special case bool since it is a subclass of int
            raise DataError(
                "Invalid input of type: 'bool'. Convert to a "
                "bytes, string, int or float first."
            )
        elif isinstance(value, (int, float)):
            value = repr(value).encode()
        elif not isinstance(value, str):
            # a value we don't know how to deal with. throw an error
            typename = type(value).__name__
>           raise DataError(
                f"Invalid input of type: '{typename}'. "
                f"Convert to a bytes, string, int or float first."
            )
E           redis.exceptions.DataError: Invalid input of type: 'NoneType'. Convert to a bytes, string, int or float first.

../../../micromamba/envs/bespokefit-test/lib/python3.10/site-packages/redis/_parsers/encoders.py:29: DataError
```
...

```
=========================== short test summary info ============================
FAILED openff/bespokefit/_tests/cli/test_cache.py::test_update_from_qcsubmit - redis.exceptions.DataError: Invalid input of type: 'NoneType'. Convert to a bytes, string, int or float first.
====== 1 failed, 378 passed, 4 skipped, 31 warnings in 216.19s (0:03:36) =======
```

j-wags:
## Description
Fix cache tests by updating the keys used to fetch cached results

## Todos

  - [x] Fix #383 
  - [x] ~Why did the keys change for BOTH RDKit AND OpenEye envs? If it were a cheminformatics thing I'd only expect one of them to start failing.~ This break was caused by the `openff-qcsubmit` 0.55 release 🤦


openff/bespokefit/_tests/cli/test_cache.py
```diff
@@ -137,17 +137,17 @@ def test_update_from_qcsubmit(redis_connection):
     # find the result in redis
     # key is generated by querying the redis_connection object as below, and
     # .hkeys appears to always return a length-1 list of bytes. The result is
-    # hardcoded here, generated Feburary 2024
+    # hardcoded here, last generated Feburary 2025
```
mattwthompson, line 140: The Wheel of Time keeps on turning

devtools/conda-envs/no_openeye.yaml
```diff
@@ -23,7 +23,7 @@ dependencies:
   - openff-utilities
   - openff-toolkit >=0.15
   - openff-forcefields >=2024.04.0
-  - openff-qcsubmit >=0.50
+  - openff-qcsubmit >=0.55
```
mattwthompson, line 26: Curious why you bumped it this far and not any more? Fine either way, just wanted to know if there was a reason I've forgotten about

mattwthompson (approved):
LGTM, just one question for my curiosity

devtools/conda-envs/no_openeye.yaml
j-wags: 0.55 is the newest version of QCSubmit (I tried to be cute with the versioning and jumped to 0.50 when the MolSSI QC* packages did, but now it's just creating confusion since QCFractal/Portal are up to 0.58 and we're going at a different release pace)

mattwthompson: I intuit that 1 is the best tradeoff of effort and value, especially since the manual updates are somewhat documented

I don't think this case hits production, otherwise we'd need to think about pinning in packaging

mattwthompson: Thanks josh!

mattwthompson:
Based on the feedback from our QCSubmit wizard - that the breaking change from QCSubmit isn't something we're likely going to need to deal with frequently - I think merging this as-is is the clear path forward.

devtools/conda-envs/no_openeye.yaml
j-wags: Ahh, that makes sense. Thanks Josh!

PR #395, opened 2025-04-22
Drop single-file installer action

### LINKED ISSUE #394: Drop installer action
**Description**

I was looking for something else and found that this project still has the "single-file installer" action. It has not been used in [almost three years](https://github.com/openforcefield/openff-bespokefit/releases/tag/0.1.2) and is now cruft.

mattwthompson:
## Description
Closes #394, and can be reverted in the future if the file is useful again


j-wags (approved):
Thanks, approved pending also deleting https://github.com/openforcefield/openff-bespokefit/blob/main/devtools/conda-envs/installer.yaml

Issue #403, opened 2025-06-18
Deployment CI is failing


j-wags:
https://github.com/openforcefield/openff-bespokefit/actions/runs/15721065224

j-wags:
This most recent tag was being correctly detected just [two days ago](https://github.com/openforcefield/openff-bespokefit/actions/runs/15680593894/job/44170881197#step:6:65)

PR #404, opened 2025-06-18
ask pretty please for deployment CI to actually fetch recent tags

j-wags:
Fix https://github.com/openforcefield/openff-bespokefit/issues/403


j-wags:
[Huzzah!](https://github.com/openforcefield/openff-bespokefit/actions/runs/15739075546)

PR #406, opened 2025-06-30
Fix #405

### LINKED ISSUE #405: Deployment CI is failing
https://github.com/openforcefield/openff-bespokefit/actions/runs/15864518162/job/44728781361

```
FAILED openff/bespokefit/_tests/executor/utilities/test_celery.py::test_get_task_information_success[celery_setup_worker] - RuntimeError: Never call result.get() within a task!
See https://docs.celeryq.dev/en/latest/userguide/tasks.html#avoid-launching-synchronous-subtasks
```

j-wags:
## Description
Minimal changes to get CI passing. See [this page](https://docs.celeryq.dev/en/latest/userguide/tasks.html#avoid-launching-synchronous-subtasks). I tried to do something fancier-and-more-correct but I didn't grok the issue well enough to do it right, see the first commit for my progress on that.

## Todos
  - [x] Fixes #405

### DIFF
--- a/openff/bespokefit/_tests/executor/utilities/test_celery.py
+++ b/openff/bespokefit/_tests/executor/utilities/test_celery.py
@@ -80,12 +80,16 @@ def test_spawn_worker(spawn_function, celery_app, monkeypatch):
 
 
 def test_get_task_information_success(celery_app, celery_worker):
+    #chain = mock_task_success.s()
+    #task_result = chain()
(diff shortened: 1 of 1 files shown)


Yoshanuikabundi (commented):
Wow. Thanks, Celery.

Possibly preferable alternative: https://github.com/openforcefield/openff-bespokefit/pull/407

If that doesn't work, then this seems to be a good solution.

PR #407, opened 2025-07-02
Alternative fix for #405

### LINKED ISSUE #405: Deployment CI is failing

Yoshanuikabundi:
## Description
Originally, `pytest-celery` seems to have been a thin wrapper around the `celery.contrib.pytest` pytest plugin distributed with celery. `pytest-celery` was just a way to load that plugin by installing a package. More recently, `pytest-celery` has added docker-based features that break in GHA. This change is described cryptically [here](https://github.com/celery/pytest-celery/blob/4d1a1dbcf340abab43d773b928f824a5a4ce7f08/README.rst#official-pytest-plugin-for-celery) and [here](https://docs.celeryq.dev/en/latest/userguide/testing.html).

Fortunately, it's possible to load `celery.contrib.pytest` directly in `conftest.py` without the `pytest-celery` package - see <https://docs.celeryq.dev/en/latest/userguide/testing.html#enabling>. Locally on my machine the celery tests appear to pass when `pytest-celery` is removed and `celery.contrib.pytest` is loaded manually. I think this approach is preferable to a downpin to an alpha version of a package.

Fixes #405.

Note that the minimal fix described above does not fix deployment tests, as they use the latest release and do not pull down the new `conftest.py`. Fortunately, (micro)mamba has recently added support for setting environment variables in conda environments! So this PR additionally enables the `celery.contrib.pytest` plugin via environment variable in the deployment environments, with a comment that it can be removed after the next release (at which point the updated `conftest.py` can take over)

## Todos
  - [x] Remove `pytest-celery` dependency from all environments
  - [x] Manually load `celery.contrib.pytest` plugin in root `conftest.py`
  - [ ] ~~Pass deployment tests: <https://github.com/openforcefield/openff-bespokefit/actions/runs/16018342209>~~
  - [x] Set temporary environment variable in deployment environments until next release
  - [x] Pass deployment tests: <https://github.com/openforcefield/openff-bespokefit/actions/runs/16018906783>

## Status
- [ ] Ready to go


j-wags (approved):
Wow, great work and a much cleaner fix. Thanks @Yoshanuikabundi!

PR #414, opened 2025-11-03
Race condition fix

jthorton:
## Description
I think there is a race condition in the coordinator worker [cycle](https://github.com/openforcefield/openff-bespokefit/blob/ef15e0dd523ff441f463833f29276217fd0fb337/openff/bespokefit/executor/services/coordinator/worker.py#L58) method. This results in the task id we get from looking at the head of the task queue [here](https://github.com/openforcefield/openff-bespokefit/blob/ef15e0dd523ff441f463833f29276217fd0fb337/openff/bespokefit/executor/services/coordinator/worker.py#L72) not matching the id we process after updating the [task](https://github.com/openforcefield/openff-bespokefit/blob/ef15e0dd523ff441f463833f29276217fd0fb337/openff/bespokefit/executor/services/coordinator/worker.py#L85) if there are few tasks in the queue this results in a `NoneTypeError` that we sometimes see in the CI or in production mode with a low number of jobs. 

This update uses a snapshot of the task queue to process the results rather than relying on the queues not changing between the peek, updating the task and then removing the id. 

## Todos
  - [ ] TODO 1

## Questions
- [ ] Question1

## Status
- [ ] Ready to go

### DIFF
--- a/openff/bespokefit/executor/services/coordinator/worker.py
+++ b/openff/bespokefit/executor/services/coordinator/worker.py
@@ -9,10 +9,11 @@ from openff.bespokefit.executor.services.coordinator.storage import (
     TaskStatus,
     get_n_tasks,
     get_task,
-    peek_task_status,
     pop_task_status,
     push_task_status,
+    remove_task_status,
     save_task,
+    snapshot_task_status,
 )
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


j-wags:
I've been trying to verify this fixes the deployment test failures, but after several hours I'm unable to reproduce the deployment test failures in the first place. I'd hoped to either reproduce these locally or in GH actions to verify the fix, but weirdly, the regular CI runs the exact same integration test (the ethane torsiondrive) as the failing deployment CI, and it passes every time. 

My furthest progress(?) in reproducing locally has been setting higher values of of `await asyncio.sleep(...)` in `executor/services/coordinator/worker.py` and lower values of `BEFLOW_COORDINATOR_MAX_UPDATE_INTERVAL` on the CLI, but this yields a different error (`[ERROR] failed to connect to the bespoke executor - please make sure one is running and your connection settings are correct`) than the error message in deployment CI (the `NoneType` return). I'll keep pushing on this today, but am open to other tips!

openff/bespokefit/executor/services/coordinator/worker.py
```diff
@@ -9,10 +9,11 @@
     TaskStatus,
     get_n_tasks,
     get_task,
-    peek_task_status,
```
j-wags: This is now only used in tests.

j-wags (changes requested):
This looks pretty good. Just one request for more tests and I'll approve this.

PR #432, opened 2026-05-27
Downpin QCElemental and QCEngine

mattwthompson:
## Description
New versions of these package use a new QCSchema which we do not currently support

## Todos
  - [x] Downpin `qcelemental <0.50`
  - [x] Downpin `qcengine <0.50`

## Status
- [x] Ready to go

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.12", "3.13"]
         pydantic-version: ["1"]
         openeye: [true, false]
         integration: [true]
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        python-version: ["3.12", "3.13"]
+        python-version: ["3.11", "3.12"]
         pydantic-version: ["1"]
         openeye: [true, false]
         integration: [true]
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macOS-latest]
-        python-version: ["3.11", "3.12"]
+        python-version: ["3.12", "3.13"]
```
j-wags: (blocking) It looks like the env can't solve with python 3.13 (something with forcebalance). This doesn't need to be propagated to the feedstock since a new forcebalance release is impending (which will likely solve this), but it is useful to see.

```suggestion
        python-version: ["3.11", "3.12"]
```

devtools/conda-envs/no_openeye.yaml
```diff
@@ -23,7 +23,8 @@ dependencies:
   - rdkit =2024.03.5
   - ambertools >=22
   - openff-utilities
-  - openff-toolkit >=0.15
+  - openff-toolkit =0.16.5
+  - openff-interchange =0.3.29
```
j-wags: (blocking) If we're not bumping python versions, we probably want to revert these as well (though I'm happy to keep an interchange pin that's reflective of the original toolkit pin)

j-wags (changes requested):
Since we're putting Bespokefit into maintenance mode, I think our actions here should be mostly fixing failing CI by adding upper pins and propagating those to the feedstock (note to self: Matt [already added](https://github.com/conda-forge/openff-bespokefit-feedstock/blob/940abb84a2480389143c2cf5acb027e968998bcc/recipe/meta.yaml#L45-L46) the QCE package downpins to the feedstock). So here I don't think we should bump the python version (ie lets keep testing on py311 and 312), and just add downpins as necessary for the QCE* packages.

devtools/conda-envs/no_openeye.yaml
mattwthompson: These aren't related to the Python version bumps

j-wags (approved):
LGTM! Thanks @mattwthompson.

PR #439, opened 2026-08-24
Vendor AmberTools provenance

mattwthompson:
Unfortunately this won't work because the toolkit [also depends on this](https://github.com/openforcefield/openff-toolkit/blob/0.19.0/openff/toolkit/utils/ambertools_wrapper.py#L75-L80) unsupported code path - it probably should just live in the toolkit period

## Description
- [x] Port fix to `get_ambertools_version`
- [x] Squash any errors around above function

## Status
- [ ] Ready to go


mattwthompson:
Doesn't work because the same failure will just propagate through the toolkit

